PR #33, opened 2026-03-19
feat: add Linux support (evdev + HID++ + uinput)

hieshima:
Add Linux platform support for Mouser, including mouse event interception, keyboard action simulation, HID++ device communication, and per-app profile detection.

Linux implementation:
- `mouse_hook.py`: evdev grab + uinput forwarding for button/scroll blocking, gesture tracking with threading lock, crash guard (atexit + signal handlers) to release grab on abnormal exit
- `key_simulator.py`: evdev UInput virtual keyboard, 22 actions with thread-safe lazy initialization
- `app_detector.py`: xdotool + /proc for foreground app detection (X11)
- `config.py`: XDG-compliant config path (~/.config/Mouser)
- `requirements.txt`: evdev>=1.6 Linux dependency

Cross-platform fixes:
- `hid_gesture.py`: support both "pip install hid" and "pip install hidapi" via _HidDeviceCompat wrapper; cap HID++ control discovery at 32 with circuit breaker to prevent DoS from malicious devices
- `config.py`: atomic config writes (tempfile + os.replace) to prevent data loss on crash; restrictive file permissions (0o600 on POSIX); type validation on config load

Note: Cross-platform changes (`config.py`, `hid_gesture.py`) were reviewed against Windows/macOS code paths but only runtime-tested on Linux (Nobara/Fedora, Bluetooth MX Master 3S).

Tested: 27/27 unit tests pass. Manual verification on Nobara Linux with Logitech MX Master 3S over Bluetooth.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -397,12 +397,20 @@ class AppCatalogTests(unittest.TestCase):
             patch.object(app_catalog.sys, "platform", "linux"),
             patch.object(app_catalog, "get_app_catalog", return_value=fake_catalog),
             patch.object(app_catalog.os.path, "exists", return_value=True),
+            patch.object(
+                app_catalog.os.path,
+                "realpath",
+                side_effect=lambda value: {
+                    "/usr/bin/firefox": "/opt/firefox/firefox",
+                    "/usr/lib64/firefox/firefox": "/usr/lib64/firefox/firefox",
+                }.get(value, value),
+            ),
         ):
             resolved = app_catalog.resolve_app_spec("/usr/lib64/firefox/firefox")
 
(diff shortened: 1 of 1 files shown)


thisislvca:
Hey, thanks so much for contributing!

I pushed a few follow-up commits directly onto this PR to add some missing features:

- added Linux app profile creation/discovery
- added regression coverage for the Linux app-profile path
- enabled gesture swipe mappings in the Linux UI
- documented current Linux runtime requirements and caveats

Before we merge, could you please test this on Linux and confirm:

- app profile creation works
- per-app auto-switching works
- gesture tap + swipe mappings work
- button remapping works with your current permissions setup

Thanks again 🙏

PR #38, opened 2026-03-22
fix: add MX Master 4 device to known logi devices

awkure:
## Summary

Registers the Logitech MX Master 4 in the known-device catalog so Mouser can resolve it by USB product ID and by common HID product strings, using the existing `mx_master` UI layout.

## Changes

- **`core/logi_devices.py`**: Add a `LogiDeviceSpec` for `mx_master_4` with product ID `0xB042`, aliases (including “MX Master 4 for Mac”, `MX_Master_4`, and “for Business”), and `ui_layout="mx_master"`.
- **`tests/test_logi_devices.py`**: Add unit tests that assert resolution by product ID and by HID-style product name.

### DIFF
--- a/core/logi_devices.py
+++ b/core/logi_devices.py
@@ -72,6 +72,18 @@ class ConnectedDeviceInfo:
 # Solaar/logiops for the major MX-family mice we want to grow into first.
 KNOWN_LOGI_DEVICES = (
     LogiDeviceSpec(
+        key="mx_master_4",
+        display_name="MX Master 4",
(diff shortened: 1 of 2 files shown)


thisislvca:
@Thxlx Put up a follow-up for this here: #42

it wires macOS button 6 into the gesture handlers in `mouse_hook.py` and adds a regression test around that path.

if you get a chance, can you try that PR on your setup and let me know if the action button works there?

PR #64, opened 2026-03-30
fix: re-enable CGEventTap on timeout and skip trackpad scroll events

hieshima:
## Problem

Two macOS CGEventTap issues that affect daily use:

1. **Button/scroll remapping silently stops working after running for a while.**
   macOS disables a `CGEventTap` when the callback responds too slowly (`kCGEventTapDisabledByTimeout`). The existing wake observer (`NSWorkspaceDidWakeNotification`) only covers sleep/wake — it does not handle runtime timeout disabling. Once disabled, all remapping permanently stops until restart.

2. **Trackpad gestures and scroll direction break while Mouser is running.**
   The `kCGEventScrollWheel` handler intercepts *all* scroll events — including trackpad two-finger scrolls — without distinguishing input source. This causes:
   - Trackpad horizontal swipes trigger mapped mouse actions (e.g., browser back/forward)
   - `invert_vscroll` applies to trackpad, overriding the system scroll direction
   - macOS gesture recognition breaks because scroll phase sequences are interrupted

Related: #35, PR #45

### 1. Handle `kCGEventTapDisabledByTimeout` / `kCGEventTapDisabledByUserInput`

At the top of `_event_tap_callback`, detect the two disabled signal event types (`0xFFFFFFFE`, `0xFFFFFFFF`) and immediately call `CGEventTapEnable(tap, True)`. This is the [Apple-recommended pattern](https://developer.apple.com/documentation/coregraphics/cgeventfiltermasksettap) for keeping an event tap alive.

### 2. Filter trackpad events via `kCGScrollWheelEventIsContinuous`

At the top of the `kCGEventScrollWheel` handler (before any dispatch/block/inversion logic), check CGEvent field 88 (`kCGScrollWheelEventIsContinuous`):
- `0` → discrete mouse wheel event → process normally (dispatch, block, invert)
- `1` → continuous trackpad/Magic Mouse event → `return cg_event` (pass through untouched)

This is the same approach used by [Scroll Reverser](https://github.com/pilotmoon/Scroll-Reverser), [Mos](https://github.com/Caldis/Mos), and [LinearMouse](https://github.com/linearmouse/linearmouse). Field 88 is a [documented Apple API](https://developer.apple.com/documentation/coregraphics/cgeventfield/kcgscrollwheeleventiscontinuous) stable since macOS 10.5.

### CGEventTap re-enable (manual test on macOS 15.5)

Added a temporary `time.sleep(5)` inside `_event_tap_callback` to force macOS to trigger the timeout mechanism. Result:

```
07:51:01 [MouseHook] CGEventTap: first event received
07:51:06 [MouseHook] CGEventTap disabled by system (type=0xFFFFFFFE), re-enabling
07:51:17 [MouseHook] CGEventTap disabled by system (type=0xFFFFFFFE), re-enabling
```

The tap was automatically re-enabled each time macOS disabled it. Remapping resumed without restart.

### Trackpad filtering (manual test on macOS 15.5)

- Trackpad two-finger horizontal swipe: system gestures work normally, no spurious actions triggered ✅
- Trackpad vertical scroll: follows system scroll direction setting (standard/natural) ✅
- Mouse wheel horizontal tilt: mapped action fires correctly ✅
- Mouse wheel vertical scroll with `invert_vscroll`: inversion applies correctly ✅

### Automated tests

```
$ uv run python -m unittest discover -s tests -v
Ran 176 tests in 0.051s
OK (skipped=28)
```

New tests added:
- `test_reenable_on_timeout` — callback receives `0xFFFFFFFE`, asserts `CGEventTapEnable(tap, True)` is called
- `test_reenable_on_user_input` — same for `0xFFFFFFFF`
- `test_normal_event_does_not_reenable` — normal events don't trigger re-enable
- `test_trackpad_scroll_passes_through_callback` — continuous scroll returns `cg_event` as-is
- `test_trackpad_hscroll_not_blocked` — trackpad horizontal scroll not dispatched as HSCROLL action
- `test_mouse_wheel_hscroll_dispatched_and_blocked` — discrete mouse wheel horizontal scroll dispatches and blocks correctly

### DIFF
--- a/core/mouse_hook.py
+++ b/core/mouse_hook.py
@@ -934,6 +934,8 @@ elif sys.platform == "darwin":
     _BTN_BACK = 3
     _BTN_FORWARD = 4
     _SCROLL_INVERT_MARKER = 0x4D4F5553
+    _kCGEventTapDisabledByTimeout = 0xFFFFFFFE
+    _kCGEventTapDisabledByUserInput = 0xFFFFFFFF
 
     class MouseHook:
         """
@@ -1301,6 +1303,15 @@ elif sys.platform == "darwin":
         def _event_tap_callback(self, proxy, event_type, cg_event, refcon):
             """CGEventTap callback.  Return the event to pass through, or None to suppress."""
             try:
+                if event_type in (
+                    _kCGEventTapDisabledByTimeout,
+                    _kCGEventTapDisabledByUserInput,
+                ):
+                    print(f"[MouseHook] CGEventTap disabled by system "
+                          f"(type=0x{event_type:X}), re-enabling", flush=True)
+                    Quartz.CGEventTapEnable(self._tap, True)
+                    return cg_event
+
                 if not self._first_event_logged:
                     self._first_event_logged = True
                     print("[MouseHook] CGEventTap: first event received", flush=True)
@@ -1379,6 +1390,13 @@ elif sys.platform == "darwin":
                         Quartz.CGEventGetIntegerValueField(
                             cg_event, Quartz.kCGEventSourceUserData
                         ) == _SCROLL_INVERT_MARKER
+                    ):
+                        return cg_event
+                    # Pass through trackpad / Magic Mouse continuous scroll
+                    # events untouched — only intercept discrete mouse wheel.
+                    _kCGScrollWheelEventIsContinuous = 88
+                    if Quartz.CGEventGetIntegerValueField(
+                        cg_event, _kCGScrollWheelEventIsContinuous
                     ):
                         return cg_event
                     h_delta = Quartz.CGEventGetIntegerValueField(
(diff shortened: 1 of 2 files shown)


TomBadash:
Thanks @hieshima -- both fixes are clean and well-documented.

Merged into master. The only conflict was a single import line in `test_mouse_hook.py` (`Mock` vs `MagicMock + call`) -- combined both sides.

The CGEventTap re-enable fix is a real quality-of-life improvement for macOS users; the trackpad filtering is the right call and it is good to see it uses the same field 88 approach as established tools like Scroll Reverser and LinearMouse.

Looking forward to the reconnect polish PR whenever you are ready.

Let us know how this holds up on your end!

TomBadash:
also if you would like to become a maintainer please leave an email for reaching out 🤝

PR #80, opened 2026-04-09
high sensitivity scroll

farfromrefug:
This has been written with the help of claude sonnet 4.6

Adds high-sensitivity (hi-res) scroll support using HID++ 2.0 features HIRES_WHEEL (0x2121) and HI_RES_SCROLLING (0x2120), following the same pending/result queue pattern used for DPI and Smart Shift.

## Protocol
- Prefers **HIRES_WHEEL (0x2121)**: read-modify-write the mode byte (bit 1 = hi-res) via functions 0x10/0x20
- Falls back to **HI_RES_SCROLLING (0x2120)**: single-byte write via function 0x10
- Setting is persisted and re-applied on device connect, alongside DPI and Smart Shift

## Changes

- **`core/hid_gesture.py`** — adds `FEAT_HI_RES_SCROLL`/`FEAT_HIRES_WHEEL` constants; discovers both features in `_try_connect`; `hi_res_scroll_supported` property; thread-safe `set_hi_res_scroll(enabled)` + `_apply_pending_hi_res_scroll()` executed on the listener thread; state reset on disconnect
- **`core/config.py`** — adds `hi_res_scroll: false` default, bumps config to version 7 with migration
- **`core/engine.py`** — `set_hi_res_scroll()`, `hi_res_scroll_supported`, applies saved value on startup
- **`ui/backend.py`** — exposes `hiResScrollSupported`, `hiResScroll` properties and `setHiResScroll(bool)` slot to QML
- **`ui/qml/ScrollPage.qml`** — "High Sensitivity Scroll" card with toggle, gated on `backend.hiResScrollSupported`
- **`tests/test_config.py`** — updated migration version assertions, added `hi_res_scroll` assertion


TomBadash:
Hey @farfromrefug, thank you for the PR, really appreciate the work here.

I tested it on Windows 11, and it definitely creates a noticeably different (and improved) scrolling experience.

One thing that feels missing from my perspective is the ability to control the sensitivity level. At the moment, it’s a bit too sensitive on my setup, and having some form of user control over the scroll intensity would likely make this feature more useful across a wider range of users and preferences.

Would you be open to addressing this in a follow-up commit?

farfromrefug:
@TomBadash yes sure I am though do you know if it is even possible?  Do you have reference/doc of some kind?
Solaar does not implement that so i am not sure

farfromrefug:
@TomBadash Could we maybe merge this as it is and try to improve it in the future? In a sense this is must have otherwise  mouse scroll is really too slow.

TomBadash:
Hi @farfromrefug,

Yes, we plan to merge it with a warning as soon as we can.

Also, I’m currently looking for Mac developers who would be interested in helping maintain the project, so we can improve review times and respond more quickly.

Would that be something you’d be open to?

farfromrefug:
@TomBadash great you'll release it.
And Ans yes sure I would be happy to help maintaon this great project you created!

Issue #81, opened 2026-04-10
[Bug] sleep/resume issue


farfromrefug:
### Mouser version

3.5.2

### Operating system

Linux

### OS version details

Ubuntu 24.04 / Macos 

### Mouse connection type

Bluetooth

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

I face the same issue i was facing with solaar
When coming back from deep sleep, sometimes the "settings" do not apply. 
It has to do with settings being applied too soon after restart.
I had a fix for solaar (which was never merged) where i was waiting a bit more to apply settings, even retrying a few times to ensure it is working
@hieshima would love your opinion on that

### Expected behavior

mouse settings applied after resume

### Steps to reproduce

1. Open Mouser
2. set settings
3. go to sleep
4. resume
5. sometimes settings are not applied

### Device info JSON

```json

```

### Log file contents

```text

```

TomBadash:
Hey @farfromrefug, could you please check v3.5.3 and let me know if it resolves your issue?

farfromrefug:
@TomBadash From my quick tests it seems to be good. I lll close it

Issue #56, opened 2026-03-27
Mapped gesture_click cannot be recognized after mouse restarted


kljjh:
After the mouse reconnects to the computer, the software cannot recognize the mapped gesture_click unless Mouser is restarted. This manifests as the corresponding button events not appearing in Debug Events. This issue occurs with both Bluetooth connections and the Logitech Unifying receiver
Device: MX Master 3

kljjh:
I noticed that version 3.5.3 attempted to address this issue. Under Bluetooth connection, the issue now appears to be resolved. After restarting the mouse, Mouser will recognize the reconnection within a few minutes, and all remappings work properly.

However, the problem still seems to persist when using the receiver. Additionally, I observed that in receiver mode, Mouser does not show the device as disconnected even when the mouse is powered off. In contrast, under Bluetooth mode, the connection is immediately terminated once the mouse is turned off.

I hope this observation can help further investigate and resolve the issue :)

TomBadash:
Hey @kljjh 

I’m really glad to hear that the issue is resolved on Bluetooth, even if only partially for now.

Regarding the USB receiver, this kind of different behavior compared to a standard Bluetooth connection has indeed been reported before. It seems like we may need to dig deeper into how we can extract more detailed device information through the receiver.

One question that could really help us narrow this down: when you’re using the USB receiver, if you turn the mouse off and then back on, do you notice any change in Mouser’s behavior? Specifically, after a minute or two, does everything start working as expected, or do you still experience issues with the functionality?

kljjh:
> Hey [@kljjh](https://github.com/kljjh)
> 
> I’m really glad to hear that the issue is resolved on Bluetooth, even if only partially for now.
> 
> Regarding the USB receiver, this kind of different behavior compared to a standard Bluetooth connection has indeed been reported before. It seems like we may need to dig deeper into how we can extract more detailed device information through the receiver.
> 
> One question that could really help us narrow this down: when you’re using the USB receiver, if you turn the mouse off and then back on, do you notice any change in Mouser’s behavior? Specifically, after a minute or two, does everything start working as expected, or do you still experience issues with the functionality?

When using the receiver, after I turn the mouse off and then back on, mouser shows no response at all. It neither indicates a disconnection nor reconnects after several minutes. Additionally, there are no corresponding disconnection or reconnection logs in the Debug Events. Aside from the buttons mentioned earlier, all other buttons can still be captured and remapped normally.

Issue #48, opened 2026-03-25
Lost mouse DPI settings


makubexhan-max:
Lost mouse DPI settings  after long standby for Win10 ，use MX master 3s

TomBadash:
Hi @makubexhan-max can you explain what do you mean and maybe attach a screenshot ?

PR #82, opened 2026-04-11
Fix reconnect replay and recover from HID transport failures

hieshima:
## Problem

After sleep, reconnect, or cold start, Mouser could get stuck in a half-recovered state:

- **Settings lost** (#81, #48): saved DPI and Smart Shift were not reliably restored because the production replay path had no settle/retry. That logic only existed in dead code (`_apply_device_settings()`, which had zero callers).
- **Gesture/mode_shift recovery gaps** (#56): diverted button callbacks could stop firing after reconnect, and there were no regressions locking this behavior down.
- **Silent transport death**: if the HID device handle went stale (for example `IOHIDDeviceSetReport failed: 0xE00002C2` on macOS), the app stayed "connected" and kept polling Smart Shift every 15 seconds instead of reconnecting.
- **Stale UI updates**: replay sent Smart Shift state as a bare string instead of the dict the backend expected, so `smartShiftChanged` could fire without actually updating config.

### Phased replay
Deleted the dead `_apply_device_settings()` helper and moved its settle/retry semantics into the live replay worker:
- **Phase A** (immediate): write saved Smart Shift so the wheel converges quickly
- **Phase B** (+3 s): re-fetch the HID handle, then write saved DPI and Smart Shift again
- **Phase C** (+5 s, only on failure): one final retry

Also moved `_emit_status()` outside `_replay_lock` to avoid callback-under-lock.

### Smart Shift serialization
Replaced the unsynchronized `_pending_smart_shift` busy-wait path with `threading.Event` plus separate call/slot locks. Reconnect cleanup now aborts pending Smart Shift requests so waiters never read stale state. Smart Shift polling is also suppressed while replay is in flight.

### Payload normalization
All engine-to-backend Smart Shift callbacks now use:
`{"mode": str, "enabled": bool, "threshold": int}`

Backend now returns early on non-dict payloads, and `smartShiftChanged` only fires after a valid update.

### Transport failure recovery
In `_request()`, active-session tx/rx exceptions now raise `IOError`, which drives the existing cleanup/reconnect path through `_main_loop()`. Discovery-time probe failures still return `None`.

This is intentionally error-code-agnostic and does not depend on any specific IOKit error code.

### Regression coverage
Added regressions for:
- replay restoring DPI and Smart Shift through the real worker path
- gesture and mode_shift callbacks firing again after reconnect
- `_divert_extras()` re-arming on reconnect and held state clearing before disconnect callbacks
- backend ignoring non-dict Smart Shift payloads without config mutation or signal emission
- Smart Shift polling being skipped during replay
- active-session transport failures raising reconnect-driving errors while discovery failures remain non-fatal

## Scope

- No config migration, no schema bump, and no new user-facing settings
- DPI still has the same pending-slot race shape, but it is deferred here because it has lower concurrency pressure
- `engine.set_smart_shift()` and `set_dpi()` remain synchronous; async dispatch is follow-up work
- `IOHIDManagerRegisterDeviceRemovalCallback` would be a more complete disconnect signal on macOS, but that is also follow-up work

## Tested on

- **macOS 15.5** (MX Master 3S over Bluetooth): full local test suite passed (`222` tests)
- **Nobara 43 KDE Wayland** (MX Master 3S over Bluetooth): full local test suite passed (`217` tests)

### DIFF
--- a/core/engine.py
+++ b/core/engine.py
@@ -456,7 +456,7 @@ class Engine:
         summary = ", ".join(f"{key}={mappings.get(key, 'none')}" for key in interesting)
         self._emit_debug(f"{prefix}: {summary}")
 
-    def _replay_saved_settings_once(self):
+    def _run_saved_settings_replay(self):
         hg = self.hook._hid_gesture
         if hg is None:
             return False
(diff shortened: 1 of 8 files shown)


TomBadash:
Hey @hieshima, I pulled this PR locally and tested the reconnect/recovery flow on my side, including disconnect/reconnect scenarios and recovery after a HID read failure. I also verified that the saved device state is restored correctly after reconnect, Smart Shift comes back, DPI is replayed correctly, and gesture handling keeps working as expected. Everything looks good from my testing.

I also want to say I’m genuinely very impressed by the quality of this PR, and honestly by all of your PRs so far. They’ve been thoughtful, well-scoped, and very helpful.

If this is something that would interest you, I’d be very happy to talk about making you a maintainer. If you want, feel free to ping me somewhere easier for real-time chat, however you prefer.

hieshima:
Hey @TomBadash, thank you for reviewing my PRs. I'd like to be a maintainer.

PR #89, opened 2026-04-12
fix: macos tray icon like other platforms

farfromrefug:
Now macos app use the same nice tray icon as linux/windows


thisislvca:
Hey! I was actually thinking of using the macOS one on all platforms, as it's easier to recognize compared to the app icon. @TomBadash and I also talked about rebranding the whole app to a cleaner color scheme/icon. Closing this for the time being :)

thisislvca:
Here are the two icons side by side for reference

image

PR #90, opened 2026-04-12
fix: macos prevent circular mapping

farfromrefug:
In my config i had:
* middle -> mode_shift
* mode_shift -> middle

This was creating a circular mapping and i could not get mode_shift -> middle to work.

Now we mark cgevents as "injected" and thus it wont try to map it again


thisislvca:
Thanks for the contribution on this one. I pushed a tiny follow-up fix for the unrelated Linux KEY_VOLUMEUP typo that was breaking CI, so the PR should be in a better spot now.

farfromrefug:
@thisislvca thanks i dont know where that KEY_VOLUMEUP change comes from. i have the same in another PR

thisislvca:
No worries, I traced it back and the `KEY_VOLUME_UP` rename was introduced in `da75644` on this PR. Before that the file still used `KEY_VOLUMEUP`, so it looks like an accidental rename :)

PR #91, opened 2026-04-12
Fix/logi bolt macos support

### LINKED ISSUE #59: MX Master 4 not detected on Windows 11
I've just installed the latest Mouser version for Windows 11 and it doesn't detect my Logitech MX Master 4.

The mouse is connected using the Logi Bolt receiver to my laptop.

My mouse is turned on, has battery and Logi+ isn't installed.

Any idea what could be causing the issue?

Image

vcanuel:
## Summary

  Fixes #59 — Logitech mice connected via a **Logi Bolt USB receiver** were not detected on macOS.

  **Root causes:**
  - The default HID backend on macOS was `iokit`-only, which skipped `hidapi` enumeration — the Bolt receiver's vendor-specific HID++ interfaces (`usage_page >= 0xFF00`) were never
  discovered
  - A `break` after failed `_divert()` stopped the device-index loop at the first non-mouse device (e.g. MX Keys S keyboard on slot 3), preventing discovery of mice on later slots

  **Changes:**
  - Default backend to `"auto"` on all platforms so `hidapi` enumerates alongside IOKit
  - On macOS, prefer IOKit for **opening** devices (non-exclusive access prevents cursor freeze), hidapi as fallback
  - Replace `break` with `continue` after divert failure so all receiver slots (1–6) are probed
  - Sort candidates so direct devices (Bluetooth) are tried before USB receivers — avoids ~10s of slot-scan timeouts for Bluetooth users
  - Show connection type ("Bluetooth" / "Logi Bolt") in the UI status badge
  - Update README + README_CN

  ## Testing

  Tested on macOS 15 (Apple Silicon) with MX Master 3S For Mac:
  - **Logi Bolt** (PID `0xC548`, receiver slot 4): all HID++ features working — REPROG_V4, DPI, Smart Shift, battery
  - **Bluetooth** (PID `0xB034`, devIdx `0xFF`): instant connection, no regression
 - Switching between modes requires an app restart (auto-reconnect on transport change is not yet supported)
  - 226 tests pass (4 new):
    - Divert failure continues to next receiver slot
    - Candidates sorted: direct devices before USB receivers
    - Transport label "Bluetooth" for devIdx 0xFF
    - Transport label "Logi Bolt" for receiver slots 1–6

  ## Screenshots

  Status badge shows **Connected · Logi Bolt** or **Connected · Bluetooth** depending on transport.

Python 2026-04-12 17 09 06

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ No telemetry. No cloud. No Logitech account required.
 - **Single instance guard** — launching a second copy brings the existing window to the front
 
 ### 🔌 Smart Connectivity
+- **Bluetooth and Logi Bolt** — works with both Bluetooth and Logi Bolt USB receivers; connection type shown in the UI
 - **Auto-reconnection** — detects power-off/on and restores full functionality without restarting
 - **Live connection status** — real-time "Connected" / "Not Connected" badge in the UI
 - **Device-aware UI** — interactive MX Master diagram with clickable hotspots; generic fallback for other models
@@ -376,7 +377,7 @@ mouser/
 │   ├── accessibility.py     # macOS Accessibility trust checks
 │   ├── engine.py            # Core engine — wires hook ↔ simulator ↔ config
 │   ├── mouse_hook.py        # Low-level mouse hook + HID++ gesture listener
-│   ├── hid_gesture.py       # HID++ 2.0 gesture button divert (Bluetooth)
+│   ├── hid_gesture.py       # HID++ 2.0 gesture button divert (Bluetooth + Logi Bolt)
 │   ├── logi_devices.py      # Known Logitech device catalog + connected-device metadata
 │   ├── device_layouts.py    # Device-family layout registry for QML overlays
 │   ├── key_simulator.py     # Platform-specific action simulator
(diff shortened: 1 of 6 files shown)


thisislvca (commented):
The connection fix looks good to me overall. One thing I want to double-check before merge is the new transport label.

Right now we label any non-`0xFF` `devIdx` as `Logi Bolt`. As far as I can tell, that only proves the mouse was reached through a receiver slot, not that the receiver itself is definitely Bolt. In other words, this seems sufficient to distinguish `Bluetooth` vs `receiver-backed`, but maybe not enough to specifically claim `Logi Bolt`.

Could you either:
- confirm that the receiver can be identified as Bolt from PID / product string / other metadata and keep the current label, or
- change the fallback label to something safer like `USB receiver` unless Bolt is positively identified?

If you tweak that part, can you also re-test both Bluetooth and the Bolt receiver once more before we merge?

vcanuel:
Thanks for the review @thisislvca ! You were right — devIdx != 0xFF only proves the device is behind a receiver slot, not that it's specifically Bolt. I checked the Solaar receiver database and confirmed that 0xC548 is the only Bolt receiver PID (notably, 0xC52B is Unifying, not Bolt).

thisislvca:
LGTM, thanks

Issue #96, opened 2026-04-13
[Device] MX Anywhere 2S


woxiqingxian:
### Device name and model

Logitech MX Anywhere 2S

### How is the device connected?

Both (I have tried both)

### Does Mouser detect the device at all?

Yes, it shows the correct model name

### Device info JSON

```json

```

### Physical buttons on your mouse

<img width="694" height="543" alt="Image" src="https://github.com/user-attachments/assets/3f39036e-350b-4b53-b287-11fcb057e588" />

### Are you willing to help test?

Yes, I can test builds and provide feedback

Issue #94, opened 2026-04-13
[Bug] MX Anywhere 2S Middle Button is 1 Button (Should be @)


pcchristie:
### Mouser version

3..3

### Operating system

Windows 11

### Mouse connection type

Bluetooth

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

When I go to re-map the Middle Button on the Anywhere 2S it shows as one "middle button."

For the A2S, the middle button can't be clicked normally ("in") but can be clicked left or right - this should therefore have two buttons on the wheel. 

As far as I can tell, there is only one in Mouser.

### Expected behavior

As above.

### Steps to reproduce

As above. Mouse is correctly auto-detected.

### Device info JSON

```json
{
  "device_key": "mx_anywhere_2s",
  "display_name": "MX Anywhere 2S",
  "product_id": "0xB01A",
  "product_name": "Wireless Mobile Mouse MX Anywhere 2S",
  "transport": "hidapi",
  "ui_layout": "mx_anywhere_2s",
  "supported_buttons": [
    "middle",
    "gesture",
    "gesture_left",
    "gesture_right",
    "gesture_up",
    "gesture_down",
    "xbutton1",
    "xbutton2"
  ],
  "gesture_cids": [
    "0x00D7"
  ],
  "dpi_range": [
    200,
    4000
  ],
  "discovered_features": {
    "REPROG_V4 (0x1B04)": "index 0x09",
    "ADJUSTABLE_DPI (0x2201)": "index 0x0A",
    "BATTERY (0x1000)": "index 0x06"
  },
  "reprog_controls": [
    {
      "index": 0,
      "cid": "0x0050",
      "task": "0x0038",
      "flags": "0x0001",
      "mapped_to": "0x0050",
      "mapping_flags": "0x0000"
    },
    {
      "index": 1,
      "cid": "0x0051",
      "task": "0x0039",
      "flags": "0x0001",
      "mapped_to": "0x0051",
      "mapping_flags": "0x0000"
    },
    {
      "index": 2,
      "cid": "0x0052",
      "task": "0x00A9",
      "flags": "0x0131",
      "mapped_to": "0x0052",
      "mapping_flags": "0x0011"
    },
    {
      "index": 3,
      "cid": "0x0053",
      "task": "0x003C",
      "flags": "0x0131",
      "mapped_to": "0x0053",
      "mapping_flags": "0x0001"
    },
    {
      "index": 4,
[...]

PR #98, opened 2026-04-13
feat: Support for Anywhere 2S model

MaxwellChang98:
#96 #94 -Support for Anywhere 2S model.
- Removed Middle Click: On this model, the middle button is a physical switch for scroll wheel friction (ratchet/freespin) and lacks electronic trigger functionality.
- Added Back and Forward mappings: Since the tilt wheel on this model supports left and right clicks, these have been mapped to Back and Forward functions.
- Verified on Windows: See the screenshot below for the implementation.
image

### DIFF
--- a/core/device_layouts.py
+++ b/core/device_layouts.py
@@ -163,16 +163,6 @@ MX_ANYWHERE_2S_LAYOUT = {
     "note": "",
     "hotspots": [
         {
-            "buttonKey": "middle",
-            "label": "Middle button",
-            "summaryType": "mapping",
-            "normX": 0.33,
-            "normY": 0.46,
-            "labelSide": "left",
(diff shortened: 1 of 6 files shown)


TomBadash:
Hay @MaxwellChang98 great work on the PR, and thanks for the contribution.

Just to avoid any confusion, could you please rename the Back and Forward button labels to Wheel Right and Wheel Left instead?

That would make it clearer that these actions are triggered by tilting the wheel right and left, rather than by dedicated back/forward buttons.

woxiqingxian:
Anywhere 2S users are anxiously awaiting the merger as soon as possible.
期待尽快合并，Anywhere 2S 用户焦急等待中。

PR #125, opened 2026-04-24
fix(macos): add @_autoreleased to CGEventTap callback to prevent ~1.4 GB memory leak

oraluben:
## Problem

Running Mouser for 2 days results in ~2.6 GB heap usage. Root cause: PyObjC bridge creates temporary ObjC objects on every CGEventTap callback and `NSWorkspace.frontmostApplication()` call. Without an explicit `NSAutoreleasePool`, these objects are never released.

Diagnosed with `heap` on a live process (PID 20922, uptime ~2 days):

| Source | Leaked objects | Count | Size |
|---|---|---|---|
| `_event_tap_callback` | HIDEvent, CGEvent, CGSEventAppendix, NSMutableArray | ~4.5M | **~1.4 GB** |
| `get_foreground_exe` | GPProcessMonitor, GPProcessInfo | ~42K | **~108 MB** |

## Fix

Add an `@_autoreleased` decorator (wraps `objc.autorelease_pool()`) to:
- `_event_tap_callback` in `mouse_hook.py` — called ~19 times/sec from mouse movement alone
- `get_foreground_exe` in `app_detector.py` — polled every 0.3s on a daemon thread

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -2,6 +2,24 @@
 
 This document contains technical details designed to help new developers understand the core components of the Mouser project.
 
+## Development Setup
+
+Install the project dependencies before running the app or test suite. Do not rely on the system Python unless it already has the requirements installed.
+
+```bash
+python3 -m venv .venv
+source .venv/bin/activate
+python -m pip install -r requirements.txt
+```
+
+On macOS, `requirements.txt` installs PyObjC (`objc`, Quartz, and AppKit bindings), which Mouser needs for CGEventTap, app detection, and key simulation.
+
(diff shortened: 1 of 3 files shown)


thisislvca:
Looks good overall, thanks so much for the fix! I pushed a small follow-up on top of the branch to make the `PyObjC` dependency behavior explicit: since this fix is macOS-specific and relies on `objc.autorelease_pool()`, missing `PyObjC` now fails with a clear setup message instead of falling into a confusing import/callback error.

Also added a short dev setup note so contributors run tests from an env with `requirements.txt` installed. Local full test suite passes for me after installing deps :)

Issue #146, opened 2026-05-03
[Bug] Cannot startup the desktop app in linux


drunkenQCat:
### Mouser version

v3.6.0

### How are you running Mouser?

Official GitHub release asset

### Build / asset details

Mouser-Linux.zip

### Operating system

Linux

### OS version details

SteamOS 3.7.22

### Mouse connection type

USB receiver (Bolt / Unifying)

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

when startup, it report error:
```bash
(deck@steamdeck Mouser)$ ./Mouser 
Traceback (most recent call last):
  File "main_qml.py", line 49, in <module>
ImportError: /home/deck/Softwares/Mouser/_internal/libQt6DBus.so.6: file too short
[PYI-147242:ERROR] Failed to execute script 'main_qml' due to unhandled exception!
```

### Expected behavior

Start up the desktop app.

### Steps to reproduce

just open it in console.

### Device info JSON

```json

```

### Log file contents

```text
empty
```

sedut-pank:
fedora 43 is the same error

Level12Lobster:
Same error on CachyOS

Nick-b404:
Same error on fedora KDE, still not merged

PR #160, opened 2026-05-13
fix: repair Linux release asset packaging

hieshima:
## Summary

- Fix the Linux release bundle so the packaged `.desktop` template, udev rules,
  and permission helper land at `_internal/linux/<file>` instead of inside
  same-named directories.
- Restore the packaged Linux launcher refresh and start-at-login enable/disable
  paths, which failed because the template file could not be read.
- Add a packaging spec invariant test to prevent the regression from recurring.

## Addresses

- #146

## Validation

- Unit tests: `python -m pytest tests/test_startup.py tests/test_build_support.py`
- Full suite: `python -m pytest`
- Manual verification on Fedora-based Linux:
  - headless startup succeeds without `Is a directory` warnings;
  - start-at-login enable creates `~/.config/autostart/io.github.tombadash.mouser.desktop`;
  - start-at-login disable removes that file with no launcher-refresh errors.

## Notes

- Does not add new Linux package formats.
- An existing non-fatal Qt Wayland plugin warning is outside the scope of this fix.

### DIFF
--- a/Mouser-linux.spec
+++ b/Mouser-linux.spec
@@ -92,15 +92,15 @@ a = Analysis(
         (os.path.join(ROOT, "images"), "images"),
         (
             os.path.join(ROOT, "packaging", "linux", "69-mouser-logitech.rules"),
-            os.path.join("linux", "69-mouser-logitech.rules"),
+            "linux",
         ),
         (
             os.path.join(ROOT, "packaging", "linux", "install-linux-permissions.sh"),
-            os.path.join("linux", "install-linux-permissions.sh"),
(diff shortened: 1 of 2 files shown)


Nick-b404:
When will the next release be? This bug completely breaks the linux app, so the most recent version doesn't work until this change is pushed into a release version. @TomBadash

Nick-b404:
This issue is also present on the previous release too, so going back a version isn't possible. I'll have a look and see where it stops but a new release is needed pretty soon.

PR #166, opened 2026-05-14
build(macos): sign hardened-runtime bundle with stable cdhash

hughesyadaddy:
## Summary

- Add an opt-in local signed macOS build path driven by `MOUSER_SIGN_IDENTITY`.
  When set, the script signs nested bundle code with hardened runtime options,
  signs `dist/Mouser.app` with `build_resources/Mouser.entitlements`, and
  verifies the result with `codesign --verify --deep --strict`.
- Default behavior is unchanged: without `MOUSER_SIGN_IDENTITY`, the script keeps
  the existing ad-hoc signing path.
- Make Python selection explicit and environment-manager agnostic. Resolution is:
  `MOUSER_PYTHON` -> active `$VIRTUAL_ENV/bin/python3|python` -> repo
  `./.venv/bin/python3|python` -> `python3|python` on `PATH`. The script does not
  call pyenv, Conda, uv, asdf, Poetry, or other managers directly.
- Fail fast when the selected interpreter is unusable or missing PyInstaller, and
  print concise build provenance: resolver source, Python path/version/arch,
  PyInstaller version, target arch when set, signing mode, and entitlements hash
  in signed mode.
- Pin `PYTHONHASHSEED=0` for PyInstaller so repeat builds with the same source,
  interpreter, dependency set, architecture, signing identity, entitlements, and
  timestamp policy are less likely to drift.
- Sign identity-mode nested `.dylib`, `.so`, and `.framework` entries with a
  depth-first pass before signing the outer app. This avoids invalidating parent
  framework signatures after nested code is signed.
- Document the local-only signing boundary in `README.md`, `README_CN.md`,
  `readme_mac_osx.md`, and `DEVELOPMENT.md`.

## Scope

This is a local developer signing convenience intended to reduce macOS
Accessibility permission churn during repeated local builds when build inputs stay
stable. It is not a Developer ID distribution, notarization, stapling, or
Gatekeeper-ready release-signing workflow. Public macOS release zips remain
ad-hoc signed until a separate release-signing workflow exists.

## Validation

- `zsh -n build_macos_app.sh`
- `plutil -lint build_resources/Mouser.entitlements`
- `python -m py_compile main_qml.py core/*.py ui/*.py`
- `python -m pytest tests/test_macos_build_script.py tests/test_build_support.py`
- `python -m pytest`
- `git diff --check`

## Notes

- Identities can be listed with `security find-identity -v -p codesigning`; SHA-1
  form is preferred because it survives keychain renames.
- pyenv users should initialize shims so `python3` resolves through pyenv, or set
  `MOUSER_PYTHON` explicitly.


hieshima:
Thanks @hughesyadaddy. The "TCC grants don't survive local rebuilds" problem is a real pain point, and the `MOUSER_SIGN_IDENTITY` + `PYTHONHASHSEED=0` direction is the right shape for local developer builds.

I pushed a follow-up commit (`edf52b2`) directly to this PR branch. Your original commit is preserved as-is; the follow-up tightens the implementation around the same design:

- Simplified Python resolution to: `MOUSER_PYTHON` → active `$VIRTUAL_ENV` → repo `.venv` → `python3|python` on `PATH`.
- Added fail-fast behavior for bad explicit/high-signal interpreters or missing PyInstaller.
- Switched identity signing to one `find -depth` pass over nested `.dylib`, `.so`, and `.framework` entries.
- Refactored the script into testable functions and added hermetic build-script tests.
- Added provenance logging for Python, PyInstaller, target arch, signing mode, and entitlements hash.
- Tightened docs to frame this as local-rebuild TCC stability, not Developer ID notarization or Gatekeeper-ready release signing.

The core design decisions from your PR remain intact: opt-in env-var signing, default ad-hoc behavior, hardened runtime options, `--timestamp=none` for local cdhash stability, and the same entitlement set.

I also validated the feature end-to-end on macOS after the follow-up:

- Default `./build_macos_app.sh` builds `dist/Mouser.app`, ad-hoc signs it, and passes `codesign --verify --deep --strict --verbose=2`.
- The opt-in `MOUSER_SIGN_IDENTITY` path was tested with a temporary self-signed codesigning identity in an isolated keychain, cleaned up afterward. The script signed nested code, signed the outer app with entitlements, and passed strict deep verification.
- Two signed builds with the same temporary identity and unchanged inputs produced the same app CDHash: `32b71cf5164f6937b0f5ea132fa01aee0213d048`.

PR #167, opened 2026-05-14
feat(ui): hide-to-tray Cmd-W / Cmd-M / Esc shortcuts on main window

hughesyadaddy:
## Why

Mouser ships with `start_minimized` defaulting to `true`, and the
`feat/macos-app-shell` work turns it into a proper menu-bar app on
macOS. That's the right model — but the moment the main window is
open, the standard keyboard shortcuts users expect from menu-bar apps
do not work:

- **Cmd-W** does nothing (or fights Qt's default close). Users assume
  it'll hide the window the way every other macOS menu-bar app
  behaves and end up quitting the app or hunting for the dock icon.
- **Cmd-M** does nothing. The window minimises into the Dock under
  the hood, leaving the tray icon orphaned.
- **Esc** dismisses popups only. The window itself stays modal-feeling
  even though the app is happy to live in the tray.

All three keystrokes already have a sensible target: route them
through the same `dismiss()` path the tray icon's "Hide" item uses.

## What changed

`ui/qml/Main.qml`:

- A `dismiss()` function consolidates the existing close paths
  (tray "Hide", window close button, etc).
- Three new `Shortcut` blocks bind **Cmd-W** (`StandardKey.Close`),
  **Cmd-M** (`Ctrl+M`), and **Esc** to `dismiss()`.
- `Shortcut.WindowShortcut` scope is used so the bindings only fire
  while the main window is the active key window — they don't
  steal Cmd-W from any modal dialog that's open over the window.
- The `Esc` shortcut is gated on `!anyDialogOpen` so it doesn't
  conflict with `Popup.CloseOnEscape` (which dismisses the popup
  first).

No behaviour change for users who don't press these keys, no new
config, no new strings.

## Test plan

- [x] Launch app. Cmd-W → window hides to tray. Tray icon stays.
- [x] Re-open via tray. Cmd-M → same.
- [x] Re-open. Esc with no dialog open → same. With the
      device-list dialog open → dialog closes, window stays.
- [x] `pytest tests/` green.

### DIFF
--- a/tests/test_main_qml_shortcuts.py
+++ b/tests/test_main_qml_shortcuts.py
@@ -1,0 +1,30 @@
+from pathlib import Path
+import unittest
+
+
+ROOT = Path(__file__).resolve().parents[1]
+MAIN_QML = (ROOT / "ui" / "qml" / "Main.qml").read_text(encoding="utf-8")
+MOUSE_PAGE_QML = (ROOT / "ui" / "qml" / "MousePage.qml").read_text(
+    encoding="utf-8"
+)
+
+
+class MainQmlShortcutGuardTests(unittest.TestCase):
+    def test_mouse_page_exposes_blocking_dialog_flag(self):
+        self.assertIn("readonly property bool hasBlockingDialog", MOUSE_PAGE_QML)
+        self.assertIn("keyCaptureDialog.visible", MOUSE_PAGE_QML)
+        self.assertIn("addAppDialog.visible", MOUSE_PAGE_QML)
+        self.assertIn("deleteDialog.visible", MOUSE_PAGE_QML)
+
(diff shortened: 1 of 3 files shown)


thisislvca:

Thanks for this. The UX idea makes sense, but I don't think the shortcut scoping is tight enough yet.

In the actual QML, Cmd-W and Cmd-M are application-wide shortcuts while the main window is visible, so they can still fire while the key-capture dialog is focused or while another modal dialog is up. That’s the part I don’t want to ship yet, because it can make config harder in exactly the flows where people are typing or dismissing dialogs.

If you tighten those to the real main-window path and explicitly gate dialog/capture states, I’m happy to take another pass.

PR #170, opened 2026-05-14
feat(devices): MX Master 4 firmware-first layout metadata

hughesyadaddy:
## Summary

Adds the catalog metadata the follow-on HID++ runtime and UI PRs depend on.
No runtime behavior changes in this commit; everything lands in the listener,
engine, and mouse-hook PRs that build on top.

### `core/logi_device_catalog.py`

- **`MX_MASTER_4_BUTTONS`** — explicit button tuple that includes `thumb_button`
  for the relocated Mouse Gesture Button (CID `0x00C3`). The MX Master 4 moves
  this physical button to the thumb area and replaces the original gesture
  surface with the larger haptic Sense Panel (CID `0x01A0`).
- **MX Master 4 catalog entry** gains five new fields:
  - `supported_buttons: MX_MASTER_4_BUTTONS`
  - `has_hires_wheel: True` / `has_thumbwheel: True` — declare HID++ feature
    presence up front so the runtime skips discovery probes on every connect.
  - `gesture_cids: (0x01A0, 0x00C3, 0x00D7)` — listener tries the Sense Panel
    first; falls back to the physical button, then to the virtual gesture CID.
  - `thumb_button_cid: 0x00C3` — the relocated small button, diverted as a
    button-only extra (no rawXY) alongside whichever CID wins the gesture role.
  - `gesture_via_sense_panel: True` — enables an OS-level `btn=6` / `BTN_TASK`
    fallback path for firmware that rejects the Sense Panel divert.
- **MX Master 3S / 3 / 2S catalog entries** gain `has_hires_wheel: True` and
  `has_thumbwheel: True` so the firmware-invert path works for those devices
  too without per-device special-casing in the listener.

### `core/logi_devices.py`

- **`LogiDeviceSpec`** gains four new optional fields: `has_hires_wheel`,
  `has_thumbwheel`, `gesture_via_sense_panel`, `thumb_button_cid`. All default
  to `False` / `None` so existing device entries are unaffected.
- **`ConnectedDeviceInfo`** mirrors those four plus runtime-state twins
  (`hires_wheel_active`, `thumbwheel_active`, `active_gesture_cid`,
  `thumb_button_via_hid`) so platform mouse hooks read state directly
  without re-walking the capability inventory inline.
- **`build_connected_device_info`** accepts the new state fields as keyword
  arguments with safe defaults; callers that do not pass them get the same
  behavior as before.
- **`_LAYOUT_BUTTONS`** maps `"mx_master_4"` to `MX_MASTER_4_BUTTONS` so
  `resolve_device` returns the correct button set for the new layout key.

### `tests/test_logi_devices.py`

Extends the existing MX Master 4 device-resolution test to assert the new
catalog fields (`gesture_via_sense_panel`, `thumb_button_cid`,
`has_hires_wheel`, `has_thumbwheel`, `gesture_cids`) are present and correct.
Adds a round-trip test for `build_connected_device_info` with the new keyword
arguments.

## Testing

```
pytest tests/test_logi_devices.py -v
```

All existing tests pass; the new assertions cover the added fields.

### DIFF
--- a/core/logi_device_catalog.py
+++ b/core/logi_device_catalog.py
@@ -27,6 +27,25 @@ MX_ANYWHERE_SMARTSHIFT_BUTTONS = (
     "mode_shift",
 )
 
+# MX Master 4 layout: the standard MX Master set plus a third thumb-area
+# button. The MX Master 4 relocates the legacy gesture button (CID 0x00C3)
+# next to back/forward and adds a haptic Sense Panel (CID 0x01A0); the
+# "thumb_button" key gives the relocated button its own UI mapping target.
+MX_MASTER_4_BUTTONS = (
+    "middle",
+    "gesture",
+    "gesture_left",
+    "gesture_right",
+    "gesture_up",
+    "gesture_down",
+    "xbutton1",
+    "xbutton2",
+    "thumb_button",
+    "hscroll_left",
+    "hscroll_right",
+    "mode_shift",
+)
+
 
 def _hotspot(
     button_key: str,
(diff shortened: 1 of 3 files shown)


thisislvca:
Thanks for laying the foundation for the MX Master 4 work. I re-audited each material change against current `master`: the adapted catalog/runtime metadata shipped in #234 via commit [`13cac0b`](https://github.com/TomBadash/Mouser/commit/13cac0b2273db4f2efcf1c2c498e126d1342e2a4) and was released in [v3.7.0](https://github.com/TomBadash/Mouser/releases/tag/v3.7.0). Current code includes the dual-CID metadata, wheel capability fields, runtime state propagation, tri-state capability resolution, CID normalization, and defensive DPI coercion. #234 intentionally replaced this branch's intermediate `thumb_button` layout with the newer physical-button model (`actions_ring` for the Sense Panel plus `gesture` for the small button).

Closing this as superseded by #234. Thank you — this stack directly informed the shipped implementation.

PR #171, opened 2026-05-14
feat(devices): MX Master 4 firmware-first HID++ runtime

hughesyadaddy:
> **Stack note:** merges cleanly on top of #170 (device metadata).
> When reviewing the diff, the catalog additions and `LogiDeviceSpec`
> extensions from #170 will already be in master; only the HID++
> listener, engine, and mouse-hook code is new here.

## Why

Two user-visible problems on the MX Master 4, and a long-standing
limitation on the entire MX Master family that this PR also fixes.

### 1. Scroll-wheel inversion is silently broken across Synergy / DeskFlow / KVM / Barrier

Today Mouser inverts the wheel by intercepting OS scroll events in its
own process and re-injecting them with the sign flipped. That works on
the *host* machine, but the inverted events never leave the host:
forwarders (Synergy / DeskFlow / KVM / Barrier / USB-over-IP) read
from the original device or the host's `IOHIDEventSystem` and send
the *un-inverted* direction to remote machines. The user's
multi-monitor / hackintosh setup scrolls one direction on the host
and the opposite direction on every secondary machine. Mouser looks
broken. The forwarder looks broken. Nobody finds the bug.

The fix is to flip the sign at the **firmware** layer (HID++ feature
`0x2121` HiRes Wheel Enhanced, `0x2150` Thumbwheel), so the device
reports the inverted direction natively and every downstream consumer
— including forwarders that read the raw HID stream — sees it
already inverted.

### 2. MX Master 4's relocated buttons don't reach the right actions

The MX Master 4 reshuffles the thumb side:

- The legacy Mouse Gesture Button (CID `0x00C3`) shrinks to a small
  pressable "Thumb button" on the front face. Solaar exposes this
  feature as `Mouse_Gesture_Button` with the `Thumb_Button` alias.
- A new haptic touch surface (CID `0x01A0`, the **Sense Panel**)
  replaces it as the gesture target. Solaar calls this feature
  `Haptic`; Logitech markets it as "Haptic Sense"; Logi Options+
  exposes its gesture bindings under the "Action Ring" software
  overlay.

Mouser today only knows about `0x00C3`, so on MX Master 4:

- Gestures bound to the gesture button fire from the *small Thumb
  button*, not the Sense Panel (which is where the user actually
  expects them).
- The small Thumb button has no separate mapping target, so users
  can't bind it to its own action.

### 3. `wheel_divert` is a kill switch

For users on devices that *appear* capable but where firmware rejects
the divert (firmware variants, USB receivers that strip HID++
features, BLE quirks), the existing OS-layer path needs to remain
reachable without a rebuild. `settings.wheel_divert: "auto" | "off"`
gives users that escape hatch from `mouser.json` with no UI churn.

### `core/hid_gesture.py` — listener

- `HidGestureListener` lists `0x01A0` first in the MX Master 4
  `gesture_cids` so it diverts the Sense Panel with rawXY. Gesture
  press / release / motion now flow over the HID++ vendor channel
  and the OS never sees `btn=6`. The small Thumb button (`0x00C3`)
  is diverted in the same `setCidReporting` round as a button-only
  extra (no rawXY) so its press / release fire
  `on_thumb_button_down/up`.
- If the Sense Panel divert is rejected the listener falls back to
  the legacy CIDs (`0x00C3`, then virtual `0x00D7`), and the platform
  mouse hook picks up the OS-level fallback path (see §`mouse_hook_*`
  below).
- Discovers `FEAT_HIRES_WHEEL_ENHANCED` (`0x2121`) and
  `FEAT_THUMB_WHEEL` (`0x2150`) on connect and exposes them as
  `has_hires_wheel` / `has_thumbwheel` on `ConnectedDeviceInfo`.
- **`request_wheel_native_invert(invert_v, invert_h, timeout_s=3.0)`**
  is a cross-thread API: the engine calls it from the Qt thread, the
  listener loop applies the device write on the HID thread via a
  pending-request slot, and the caller blocks on a result event with
  a 3 s timeout. `_abort_pending_wheel_divert` wakes the waiter with
  `result=False` if the connection drops mid-request, so callers
  never strand the Qt thread waiting on a dead device.

### `core/engine.py` — engine

- **`_apply_wheel_invert_setting`** drives the divert on every
  profile / device change. When the device acknowledges, the engine
  and the platform mouse hook both flip
  `wheel_native_invert_active`; the hook then suppresses its
  OS-layer inversion path so the firmware sign and the OS layer
  don't double-invert.

### `core/config.py` — config + migrations

- Adds `thumb_button` to `BUTTON_NAMES` and `BUTTON_TO_EVENTS`.
- `DEFAULT_CONFIG` version bumps `9` → `11`.
- Migration **v10** sets `settings.wheel_divert` default (`"auto"`)
  on existing configs.
- Migration **v11** adds a default `thumb_button` mapping so the new
  button has a sensible binding out of the box on configs that
  predate this PR.

### `ui/locale_manager.py` — Chinese translations for `"Thumb button"`

Adds `"Thumb button"` → `拇指键` (zh_CN) and `拇指鍵` (zh_TW) to the
`_BUTTON_TR` table that already covers `"Middle button"`,
`"Back button"`, `"Forward button"`, etc. Without this, every
Chinese-locale user would see the English `"Thumb button"` next to
their localised mappings the moment they connect an MX Master 4.

### `core/mouse_hook_*` — platform hooks

- **`mouse_hook_contract.py`**: adds `wheel_native_invert_active` to
  the `Protocol`, so the engine can read the active path from any
  platform implementation.
- **`mouse_hook_macos.py`**: routes `btn=6` `OtherMouseDown` / `Up`
  to `THUMB_BUTTON` events and to `_begin_gesture_capture` /
  `_end_gesture_capture`. The HID-first path swallows `btn=6` while
  the HID divert holds; the OS-fallback path takes over when it
  doesn't. State mutations are protected by a new `_gesture_lock`
  that covers both the CGEventTap main-thread callback and the HID
  listener background thread (previously these mutated the same
  flags from two threads with no synchronisation).
- **`mouse_hook_linux.py`**: adds a `BTN_TASK` branch with the same
  routing.
- **`mouse_hook_windows.py`**: guards `WM_MOUSEWHEEL` /
  `WM_MOUSEHWHEEL` injection on `wheel_native_invert_active` so the
  Windows hook doesn't double-invert when firmware is doing the
  work.

## Naming (and why the comments use specific names)

The MX Master 4 thumb hardware has accumulated three different names
across Logitech's own properties, plus a fourth in the open-source
literature. Comments and identifiers in this PR pick one consistent
name per surface and acknowledge the others where it disambiguates
for readers familiar with Logitech's tooling:

| Hardware | Solaar feature | Logitech marketing | Logi Options+ UI | Used in this PR |
|---|---|---|---|---|
| CID `0x01A0`, the touch surface on the thumb side | `Haptic` | "Haptic Sense" | "Action Ring" overlay (software-level virtual ring) | **Sense Panel** |
| CID `0x00C3`, the small physical button on the front face | `Mouse_Gesture_Button` (alias `Thumb_Button` on MX Master 4) | (no marketing name) | (bound to "Mouse Gesture" actions) | **Thumb button** |

Comments call out "Action Ring" wherever it disambiguates for Logi
Options+ users, since that's the name they'll search for. They do
**not** adopt "Action Ring" as the primary identifier because it
refers to a software overlay layered on top of the Sense Panel, not
the panel itself.

## Test plan

- [x] `pytest tests/` -- **461 tests, 157 subtests, green on macOS**.
- [x] Manual: MX Master 4 over USB-C receiver on macOS. Sense Panel
      gestures drive the gesture button mapping via HID++ rawXY.
      Small Thumb button drives the `thumb_button` mapping. `btn=6`
      does not leak as a click while either path is active.
- [x] Manual: wheel scroll inversion applies before Synergy /
      DeskFlow / KVM forwarding (verified at the source machine; the
      remote machine sees the inverted direction natively).
- [x] Manual: toggling `settings.wheel_divert` from `"auto"` to
      `"off"` in `mouser.json` falls back to OS-layer inversion on
      the next profile reload (without a Mouser restart).
- [x] Manual: pulling the USB receiver mid-divert request does not
      hang the Qt thread — `_abort_pending_wheel_divert` returns
      `result=False` within `timeout_s`.
- [x] Manual: MX Master 3S regression — gestures and wheel inversion
      continue to work via the legacy CID path; `wheel_divert: auto`
      successfully drives `0x2121` on hardware that supports it.

## Cross-PR coordination with #185

Two of the commits in this PR pre-position for #185 (``fix(hooks): pass through OS mouse events when no Logitech is connected``):

- ``refactor(hooks): unify "no Logitech connected" gate via _should_intercept_events`` — hoists the inline ``self._connected_device is not None`` check from ``_apply_vscroll_invert_fallback`` / ``_apply_hscroll_invert_fallback`` into a ``BaseMouseHook._should_intercept_events`` helper. #185 also adds the same helper (it has to, to be self-contained against current master). Whichever of the two PRs merges second resolves a trivial 21-line duplicate-add of the same method definition; behavior is preserved either way.
- ``test(hooks): pin _connected_device on MacOSThumbButtonTests for #185 integration`` — a no-op against current master (this branch's CGEventTap does not yet consult ``_should_intercept_events`` at the top of the callback), but becomes load-bearing once #185 lands: without the stub device, #185's early-return short-circuits ``btn=6`` before it can reach the thumb-button dispatch path and the THUMB_BUTTON_DOWN / THUMB_BUTTON_UP / XBUTTON1_DOWN assertions start failing.


hughesyadaddy:
Pushed two follow-up fixes found during a review pass of this branch:

1. **fix(hid): stop native wheel invert from claiming success it cannot deliver** -- absent 0x2121/0x2150 features no longer count as no-op success when inversion was actually requested for that axis (a hires-wheel-only device would otherwise suppress the OS fallback and silently lose horizontal inversion), and a partial apply (vertical acked, horizontal failed) now rolls back the applied axis so the OS fallback cannot double-invert. Three new regression tests in `tests/test_wheel_divert.py`.

2. **fix(hooks): port the MX Master 4 Sense Panel fallback routing to Windows** -- macOS/Linux already swap roles when the 0x01A0 divert is rejected; Windows still used the generic handlers. Since `WH_MOUSE_LL` cannot see buttons beyond XBUTTON2, the port rides the existing Raw Input extra-button mask (the same mechanism the legacy no-HID++ path uses): extra-button transitions drive gesture capture, raw mouse motion feeds the swipe detector while the panel is held, and HID++ rawXY is suppressed during fallback. Needs on-hardware verification from a Windows + MX Master 4 owner -- specifically that the panel surfaces in `RAWMOUSE.ulRawButtons` (bit 0x20) when the firmware rejects the divert.

Full suite on this branch: 582 passed, 1 skipped, 172 subtests passed.

TomBadash:

PR #173, opened 2026-05-14
feat(ui): per-device hotspot labels in the mappings list

hughesyadaddy:
## Why

Today the mappings list shows the generic `BUTTON_NAMES` entry
("Gesture button", "Horizontal scroll left", etc.) for every device.
The MX Master 4 work in #170 adds per-device hotspot labels to the
catalog so the diagram can call out *exactly* what each button is on
that hardware (*Sense Panel*, *Top thumb button*, …), but the
mappings list on the same page is still rendering the abstract
"Gesture button" / "Thumb button" labels.

End result: the diagram says "Sense Panel", the row below it for the
same button says "Gesture button", and users have to mentally map
between the two. On MX Master 4 specifically this is confusing
because Logi Options+ also uses different names again ("Action Ring"
software overlay for the Sense Panel), so we have three names for
the same physical surface.

This PR makes the mappings list and per-app profile list pull their
labels from the same catalog source the diagram uses, so the names
match end-to-end.

## What changed

`ui/backend.py`:

- New helper `_button_label(key, fallback_name)` looks up the
  connected device layout's hotspot label first and falls back to the
  global `BUTTON_NAMES` entry when no per-device hotspot exists for
  the key.
- Both `buttons` (mappings list) and `profileButtons` (per-app
  profile mappings) route through the helper so the displayed label
  matches the diagram.
- Older MX Master mice (3S, 3, 2S, classic) have no per-device
  hotspot override for the gesture button, so they continue to
  display the global *Gesture button* entry from `BUTTON_NAMES` —
  zero regression for existing users.

## Test plan

- [x] `BackendDeviceLayoutTests.test_mx_master_4_mappings_use_per_device_hotspot_labels`:
      asserts `gesture` → `"Sense Panel"` and `thumb_button` →
      `"Top thumb button"` (per-device labels win).
- [x] `BackendDeviceLayoutTests.test_mappings_fall_back_to_button_names_when_layout_has_no_hotspot`:
      asserts the fallback path still picks up
      `"Horizontal scroll right"` from `BUTTON_NAMES` when the layout
      has no hotspot for that key.
- [x] `pytest tests/` green (463 tests, 157 subtests).

## Screenshots

MX Master 4 mappings list -- *Sense Panel* and *Top thumb button*
labels now match the diagram above them:

MX4 labels

MX Master 3S (and any other device without per-device hotspot
overrides) -- still shows the existing global *Gesture button* /
*Horizontal scroll* labels, no regression:

MX3S fallback

## Notes

- Builds on the catalog hotspot labels and the `thumb_button`
  `BUTTON_NAMES` entry from #170 / #171. Until those land, the diff
  here includes their content as well.


farfromrefug:
@hughesyadaddy i see other changes which do not seem related in this PR like addition of WHEEL_DIVERT_OFF and related code

thisislvca:
Thanks for addressing the label mismatch. The UI shipped in #234 replaced the affected flow: interactive device layouts now render their hotspot labels directly, and the MX Master 4 is modeled with separate **Actions Ring** and **Gesture button** hotspots. The generic `backend.buttons` list is now only displayed for non-interactive fallback layouts, where there is no per-device hotspot label to reconcile.

This branch is stacked, but its inherited wheel-state badge scope remains independently tracked in #172, so closing #173 does not discard that work. Closing this PR as superseded by #234, released in [v3.7.0](https://github.com/TomBadash/Mouser/releases/tag/v3.7.0).

PR #181, opened 2026-05-18
perf(ui): memoize Backend list properties to avoid per-paint rebuilds

hughesyadaddy:
## Summary

The five list-typed ``@Property`` getters on ``Backend`` — ``buttons``, ``profiles``, ``knownApps``, ``actionCategories``, ``allActions`` — are read by QML bindings inside delegate rebuilds. Every paint of the mappings list, the profile selector, or the action picker rebuilt the corresponding list from scratch, which in the case of ``profiles`` meant re-resolving every profile's apps through ``get_icon_for_exe`` and ``app_catalog.get_app_label`` on every frame.

## Approach

Each getter returns a cached snapshot. The cache is invalidated by the property's notify signal — plus any other signal whose emission actually changes the list's contents.

| Property         | Invalidating signals                            |
| ---------------- | ----------------------------------------------- |
| ``buttons``          | ``mappingsChanged``, ``deviceLayoutChanged``        |
| ``profiles``         | ``profilesChanged``, ``activeProfileChanged``       |
| ``knownApps``        | ``knownAppsChanged``                                |
| ``actionCategories`` | ``deviceLayoutChanged``                             |
| ``allActions``       | ``deviceLayoutChanged``                             |

The notify connections are wired once in ``Backend.__init__``. The public signal contract is unchanged — every signal still fires exactly when it did before; we just stop recomputing when nothing has changed.

Each getter is split into an internal ``_compute_X()`` helper so the cache wrapper stays trivial and the original computation logic stays untouched (no behavioral risk).

## Test plan

New ``BackendListPropertyMemoizationTests`` in ``tests/test_backend.py`` (7 cases):

- [x] consecutive reads of ``buttons`` return the same object (memoized);
- [x] ``mappingsChanged`` invalidates ``buttons``;
- [x] ``deviceLayoutChanged`` invalidates ``buttons``, ``actionCategories``, ``allActions``;
- [x] ``activeProfileChanged`` invalidates ``profiles``;
- [x] ``knownAppsChanged`` invalidates ``knownApps``;
- [x] ``profiles`` cache survives ``knownAppsChanged`` and ``mappingsChanged``;
- [x] ``knownApps`` cache survives ``profilesChanged``, ``mappingsChanged``, ``deviceLayoutChanged``;
- [x] full ``pytest tests/ -q`` — 502 passed, 1 skipped, 170 subtests passed.

## Notes

- No new dependency on profiling — the rebuild cost was visible in the code itself (per-paint catalog walks and icon resolution). The caches are pure correctness-preserving memoization, not speculative optimization.
- The ``_compute_*`` split is deliberate: it keeps the diff readable as "wrap each getter" and makes the rebuild paths easy to test in isolation.

### DIFF
--- a/tests/test_backend.py
+++ b/tests/test_backend.py
@@ -1273,5 +1273,74 @@ class BackendLoginStartupTests(unittest.TestCase):
         self.assertFalse(backend.startMinimized)
 
 
+@unittest.skipIf(Backend is None, "PySide6 not installed in test environment")
+class BackendListPropertyMemoizationTests(unittest.TestCase):
+    """The five ``@Property(list, ...)`` getters on ``Backend`` (``buttons``,
+    ``profiles``, ``knownApps``, ``actionCategories``, ``allActions``) are
+    read by every QML binding that depends on them, including those evaluated
+    inside delegate rebuilds. Without memoization the lists -- and their
+    per-entry catalog / icon lookups -- were rebuilt on every paint."""
+
+    def _build(self, cfg=None):
+        loaded = copy.deepcopy(cfg or DEFAULT_CONFIG)
+        with (
+            patch("ui.backend.load_config", return_value=loaded),
+            patch("ui.backend.save_config"),
+            patch("ui.backend.supports_login_startup", return_value=False),
+        ):
+            return Backend(engine=None)
+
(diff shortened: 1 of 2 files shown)


TomBadash:

PR #182, opened 2026-05-18
fix(a11y): add Accessible roles and localize remaining hardcoded QML labels

hughesyadaddy:
## Summary

Three accessibility gaps on the Scroll page and the About dialog. None of them are visible to a sighted English user, which is why they slipped through previous reviews — but they each silently regress non-English locales or assistive technology.

## Changes

**Accessible roles on rectangle-based "buttons"**

The SmartShift mode picker (Ratchet / Free-spin) and the language selector are ``Rectangle + MouseArea`` rather than ``Button``, so screen readers previously announced them as nameless rectangles. Add:

- ``Accessible.role: Accessible.Button``
- ``Accessible.name`` bound to the visible localized text
- ``Accessible.checkable: true`` + ``Accessible.checked`` mirroring the active state
- ``Accessible.onPressAction`` that calls the same setter the ``MouseArea`` already invokes (so keyboard / AT activation works without duplicating logic)

The About dialog close ``Rectangle`` follows the same pattern.

**SmartShift toggle label localized**

``smartShiftToggle.Accessible.name`` was the literal string ``"SmartShift"``. Bind it to ``s["scroll.smart_shift"]`` so VoiceOver / Narrator / Orca speak the localized name in zh_CN and zh_TW.

**Chinese trackpad strings filled in**

``scroll.ignore_trackpad`` and ``scroll.ignore_trackpad_desc`` had English strings in both ``zh_CN`` and ``zh_TW``. Translated to match the surrounding rows.

## Test plan

New ``AccessibilityLocaleTests`` in ``tests/test_locale_manager.py``:

- [x] ``test_required_accessibility_keys_present_in_all_locales`` — ``dialog.close``, ``scroll.ignore_trackpad``, ``scroll.ignore_trackpad_desc``, ``scroll.smart_shift`` are present and non-empty in every locale.
- [x] ``test_chinese_locales_do_not_passthrough_english`` — pins that the two trackpad strings no longer ship the English fallback.
- [x] Full ``pytest tests/ -q`` — 497 passed, 1 skipped, 175 subtests passed.

## Notes

- New ``dialog.close`` key added across all three locales for reuse by future dialog chrome.
- QML changes are additive — no behavioral change for the ``MouseArea`` path, so sighted-mouse users see the same UI.


thisislvca:
tiny follow-up from my side: I pushed a fix for the missing `Main.qml` locale binding, translated the Chinese `SmartShift` accessibility labels so they’re not still shipping English, and added a small regression check for that. if CI comes back clean I’m good merging this.

PR #185, opened 2026-05-19
fix(hooks): pass through OS mouse events when no Logitech is connected

hughesyadaddy:
## Summary

Mouser is a Logitech-mouse remapper, but the macOS ``CGEventTap`` and the Windows ``WH_MOUSE_LL`` hook are both **global** — they see events from every input device the OS knows about. Without a "is there even a Logitech attached?" gate Mouser will happily:

- swallow an ``xbutton`` click from a generic USB mouse and route it through whatever action the user mapped (e.g. ``browser_back``),
- invert a trackpad scroll because the user enabled the ``invert_vscroll`` toggle for their Logitech wheel,
- run a gesture-detection swipe over events from a mouse Mouser cannot see over HID++.

The practical user-facing failure is the **KVM scenario**: the user switches their KVM to another host, the Logitech is fully detached from this machine, but Mouser keeps running on this side and silently remaps whatever other mouse is plugged in. The same failure mode bites users in the disconnect window between sleep and HID++ reconnect.

## Fix

Add ``BaseMouseHook._should_intercept_events()`` that returns ``True`` only when ``self._connected_device is not None``. That flag flips under both HID++ connect / disconnect on macOS / Windows and under Linux evdev attach / release, so the contract is platform-uniform.

Both the macOS event-tap callback and the Windows LL-hook callback early-return the original event when the gate is closed — immediately after the existing injected-event filter and before any blocking / remapping / dispatching code. The Linux hook is already gated by construction (its evdev hook only attaches once a Logitech source device has been resolved), but consults the same helper defensively.

### Behavior under each state

| State | Hook behavior |
| --- | --- |
| No Logitech ever connected | pass-through |
| KVM points away from this host (Logitech disconnects) | next event passes through |
| KVM points back to this host (Logitech reconnects) | next event is intercepted |
| Logitech connected normally | unchanged |

The gate is evaluated per-event against live state, so transitions are instant — no event-loop tick required.

## Test plan

- [x] ``BaseMouseHookRuntimeStateTests`` gains three cases pinning the default ``False``, the flip on ``_on_hid_connect``, and the flip back on ``_on_hid_disconnect``.
- [x] ``MacOSPassthroughWhenNoDeviceTests`` exercises the real ``MouseHook._event_tap_callback`` and pins the user-visible failure modes:
  - scroll passes through when no Logitech,
  - xbutton passes through when no Logitech,
  - intercept resumes on the very next event after ``_on_hid_connect``.
- [x] ``MacOSTrackpadScrollFilterTests._make_hook`` updated to assign a stub ``_connected_device`` so the existing cases still exercise the path they mean to.
- [x] Full ``pytest tests/ -q``: 501 passed, 1 skipped, 170 subtests passed.

## Why no opt-out?

Mouser's purpose statement in the README is unambiguously about Logitech mice. A user without a Logitech bound to the host gets no value from the OS-level remap pipeline running anyway, and the previous behavior was a silent correctness bug for every KVM / multi-host user. If a future use case needs "remap any mouse" the gate can be relaxed behind an explicit setting, but the default has to be "stay out of the way".

## Cross-PR coordination with #171

#171 (``feat(devices): MX Master 4 firmware-first HID++ runtime``) also defines ``BaseMouseHook._should_intercept_events`` — identically — and routes its own OS-layer scroll-invert fallbacks (``_apply_vscroll_invert_fallback`` / ``_apply_hscroll_invert_fallback``) through that single helper instead of duplicating the ``self._connected_device is not None`` predicate.

Whichever of #171 / #185 merges second resolves a trivial 21-line duplicate-add on the same method definition. The method body is identical between the two PRs; behavior is preserved either way. #171 also pre-positions ``MacOSThumbButtonTests._make_hook`` to set ``_connected_device`` so this PR's top-level early-return does not short-circuit that test class once both PRs are integrated.

### DIFF
--- a/core/mouse_hook_base.py
+++ b/core/mouse_hook_base.py
@@ -123,6 +123,27 @@ class BaseMouseHook:
             connected_device=self._connected_device,
         )
 
+    def _should_intercept_events(self) -> bool:
+        """True only when the platform hook should block, remap, or dispatch
+        OS-level mouse events to the engine.
+
+        Mouser exists to remap a Logitech mouse's buttons. The global event
+        taps on macOS (CGEventTap) and Windows (WH_MOUSE_LL) see events
+        from every input device the OS knows about -- when no Logitech is
+        currently bound to this host (KVM switched to another machine,
+        the device is mid-reconnect after sleep, or the user simply has
+        not plugged one in) those hooks must stay completely out of the
+        way, otherwise xbutton clicks and scroll events from a trackpad
+        or generic USB mouse get swallowed and routed through Mouser's
+        remap pipeline.
+
+        Linux's evdev hook only attaches once a Logitech source device
+        has been resolved, so it is naturally gated -- but consult this
+        property defensively before dispatching there as well so the
+        contract stays platform-uniform.
+        """
+        return self._connected_device is not None
+
     def dump_device_info(self):
         hg = getattr(self, "_hid_gesture", None)
         if hg and hasattr(hg, "dump_device_info"):
--- a/core/mouse_hook_macos.py
+++ b/core/mouse_hook_macos.py
@@ -283,6 +283,18 @@ class MouseHook(BaseMouseHook):
                     return cg_event
             except Exception:
                 pass
+
+            # KVM / cold-start guard: when no Logitech is currently bound to
+            # this host, the CGEventTap must be a complete pass-through. The
+            # tap sees events from every mouse the OS knows about, so without
+            # this guard a trackpad swipe or a generic USB mouse's xbutton
+            # click would get routed through Mouser's remap pipeline -- the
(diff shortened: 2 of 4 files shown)


thisislvca:

Thanks for putting this together. I agree with the pass-through goal when no Logitech is actually bound to the host.

The thing I’m worried about is the signal you’re gating on: `_should_intercept_events()` keys off `_connected_device`, and on macOS/Windows that only flips after the HID listener finishes connecting. So on cold start, slow Bluetooth, or slow receiver enumeration, a Logitech that is already attached can still spend an initial window in pass-through mode and miss remaps until HID++ comes up.

If we can seed that state earlier or tighten the hook/listener ordering, I’m happy to re-check.

TomBadash:

PR #186, opened 2026-05-23
feat: add Homebrew cask support

### LINKED ISSUE #29: Add brew installation support
Would be great to have brew installation support. There can be multiple ways. 
1. Create tap directly from git repo https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap
2. Or add directly to homebrew https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request

dbydd:
## Summary
- add an in-repo Homebrew cask for Mouser on macOS
- document brew installation in the main README and macOS guide
- add a GitHub Actions workflow plus helper script to keep the cask version aligned with releases

## Validation
- `brew style Casks/mouser.rb`
- `python3 -m py_compile scripts/update_homebrew_cask.py`
- `python3 scripts/update_homebrew_cask.py --check --event-path /tmp/mouser-release.json`
- `python3 scripts/update_homebrew_cask.py --check --event-path /tmp/mouser-tag-only.json`

Closes #29

### DIFF
--- a/.github/workflows/update-homebrew-cask.yml
+++ b/.github/workflows/update-homebrew-cask.yml
@@ -1,0 +1,64 @@
+name: Update Homebrew cask on release
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to sync (defaults to latest release)"
+        required: false
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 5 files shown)


farfromrefug:
@dbydd the cask update process seems really complicated. As you only need to update the release we should simply add a step in the release workflow to regexp replace the version with the one we are publishing.
Plus it would be better to have that all handled in one workflow: the release workflow

PR #191, opened 2026-05-30
feat: translate Shift+wheel into horizontal scroll on macOS/Windows/Linux

FunJim:
## Summary

Hold Shift while turning the mouse wheel to scroll horizontally in every app — on macOS, Windows, and Linux. Each platform implements the translation in its own mouse hook so the behavior is consistent across the trio.

## Why

The OS doesn't translate Shift+wheel to horizontal scroll system-wide — only some apps (mostly browsers) do it themselves. With Mouser active, the user-expected "Shift+wheel = horizontal scroll" behavior is now consistent across every app.

## Behavior

- Only triggers on a real vertical wheel event with Shift held — a hardware horizontal tilt-wheel still goes through the existing `HSCROLL_LEFT/RIGHT` dispatch path untouched.
- Trackpad / continuous scroll events are not affected.
- Direction is controlled by the existing **Invert horizontal scroll** setting (`invert_hscroll`), so no new UI is required.

### macOS (`core/mouse_hook_macos.py`)

- New `_SHIFT_WHEEL_HSCROLL_MARKER` so the synthetic horizontal scroll we post passes through our own CGEventTap on re-entry without being reprocessed.
- New `_post_shift_hscroll_event` helper copies axis-1 (vertical) deltas onto axis-2 (horizontal), zeros axis-1, copies scroll/momentum phases, and strips the Shift modifier from the new event's flags so apps that already do their own Shift+scroll translation (browsers, NSScrollView) don't double-translate.

### Windows (`core/mouse_hook_windows.py`)

- New `WM_APP_INJECT_SHIFT_HSCROLL` deferred-injection message, mirroring the existing `WM_APP_INJECT_HSCROLL` invert path.
- `GetAsyncKeyState(VK_SHIFT)` is queried in the LL hook for every `WM_MOUSEWHEEL`; if Shift is held the original is blocked and a `MOUSEEVENTF_HWHEEL` is queued for the raw-input window thread to inject via `SendInput`.

### Linux (`core/mouse_hook_linux.py`)

- Opens keyboard evdev devices (read-only, no grab) on first use to query `active_keys()` for `KEY_LEFTSHIFT` / `KEY_RIGHTSHIFT`.
- In `_handle_rel`, when Shift is held during `REL_WHEEL`, writes a translated `REL_HWHEEL` on the virtual mouse and swallows the original vertical scroll; matching `REL_WHEEL_HI_RES` events are also suppressed so vertical scroll doesn't double-fire.
- The virtual uinput device always exposes `REL_HWHEEL` so the translation works even on basic mice without a tilt wheel.

## Test plan

- [x] `python -m py_compile main_qml.py core/*.py ui/*.py` — passes
- [x] `python -m unittest discover -s tests -p "test_*.py"` — 510 passed, 1 skipped (existing skip)
- [x] New `MacOSShiftWheelHScrollTests` (7 cases) — translation, Shift stripping, marker passthrough, no-Shift / tilt-wheel no-ops, `invert_hscroll` both directions.
- [x] New `LinuxShiftWheelHScrollTests` (7 cases) — keyboard discovery, `_shift_held` state, REL_WHEEL→REL_HWHEEL translation, passthrough without Shift, `invert_hscroll` direction, hi-res suppression, virtual-uinput REL_HWHEEL injection.
- [x] Manual macOS verification by the author: Shift+wheel scrolls horizontally system-wide; toggling **Invert horizontal scroll** flips the direction.
- [ ] Windows / Linux manual verification — the Windows hook lacks a unit-test harness in this repo (existing pattern), so the change is best confirmed on hardware before merging.

## Notes

- No config schema change.
- Keyboard discovery on Linux is one-shot and best-effort; if a keyboard is hot-plugged after start, it won't participate in Shift detection until the hook restarts. Acceptable for a first pass.

### DIFF
--- a/core/mouse_hook_linux.py
+++ b/core/mouse_hook_linux.py
@@ -122,6 +122,8 @@ class MouseHook(BaseMouseHook):
         self._evdev_grabbed = False
         self._evdev_remap_ready = False
         self._evdev_remap_status_state = (False, None)
+        self._keyboard_devices = []
+        self._keyboard_devices_initialised = False
 
     @property
     def evdev_ready(self):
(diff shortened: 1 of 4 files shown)


TomBadash:
Thanks @FunJim! This was integrated into the Logitech MX Master 4 support that just shipped in v3.7.0 (merged via #234), with some hybrid/adjusted merges. You're credited in the README acknowledgments. Closing as integrated, and I really appreciate the contribution. If you spot anything off in how it landed, please let me know.

PR #193, opened 2026-05-31
fix(config): preserve symlinked config.json on atomic save

FunJim:
## Summary
- When `~/Library/Application Support/Mouser/config.json` (or the Linux equivalent) is a symlink — typically managed via GNU stow into a dotfiles repo — `save_config` was clobbering the symlink with a regular file on every save.
- Root cause: `os.replace(tmp_path, CONFIG_FILE)` calls `rename(2)`, which by POSIX semantics replaces the symlink itself rather than following it.
- Fix: resolve `CONFIG_FILE` via `os.path.realpath()` and place the tempfile in the resolved target's directory, so the atomic rename updates the link's target in place. For regular files the realpath equals the original path, so behavior is unchanged.

## Test plan
- [x] `python -m unittest discover -s tests` — 499 tests pass
- [x] New tests in `tests/test_config.py` cover regular file, stowed symlink, and dangling symlink cases
- [x] Manually verified with a GNU stow-managed config on macOS

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -176,7 +176,12 @@ def load_config():
 def save_config(cfg):
     """Persist config to disk via atomic write with restrictive permissions."""
     ensure_config_dir()
-    fd, tmp_path = tempfile.mkstemp(suffix=".tmp", dir=CONFIG_DIR)
+    # Resolve symlinks so an atomic rename updates the link's target rather than
+    # clobbering the link itself — preserves setups like GNU stow that symlink
+    # config.json into a dotfiles repo.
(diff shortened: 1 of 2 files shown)


TomBadash:
Thanks @FunJim! This was integrated into the Logitech MX Master 4 support that just shipped in v3.7.0 (merged via #234), with some hybrid/adjusted merges. You're credited in the README acknowledgments. Closing as integrated, and I really appreciate the contribution. If you spot anything off in how it landed, please let me know.

PR #196, opened 2026-06-01
fix(app-detection): match macOS nested app identities

chenqianhe:
Fix macOS per-app profile matching for apps that launch nested or helper app bundles.

Previously, Mouser matched the foreground macOS app using a single identifier. Some apps launch secondary `.app` bundles from inside the main application bundle, which caused profile matching to miss the configured host app profile. This PR changes foreground app detection to return an ordered tuple of app identities, from the most specific nested app to broader host app identities, so profile lookup can first match a nested app profile and then fall back to the host app profile.

## Changes

- Return ordered foreground app identities from macOS app detection.
- Include bundle identifiers, bundle paths, bundle names, executable paths, and localized names as matching candidates.
- Update profile lookup to match ordered identities and cache alias resolution per lookup.
- Remove legacy single-string foreground/profile matching APIs.
- Tighten typing around app identity handling.
- Improve app detector tests so platform-specific imports do not leak between tests.
- Add coverage for nested macOS app identity matching and fallback behavior.

### DIFF
--- a/core/app_detector.py
+++ b/core/app_detector.py
@@ -12,7 +12,7 @@ import sys
 import threading
 import time
 
-AppIdentity = str | tuple[str, ...]
+AppIdentity = tuple[str, ...]
 
 
 def _path_from_nsurl(url) -> str | None:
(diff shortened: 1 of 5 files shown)


farfromrefug:
@chenqianhe can you provide an example app we can test with?

chenqianhe:
> @chenqianhe can you provide an example app we can test with?

I originally ran into this with a game platform app, but that may not be a convenient test case for everyone. I asked AI to help find some more common examples of real nested macOS apps that should be easier to test, such as Xcode Simulator or Instruments.

The key case is a focusable app whose .app bundle lives inside another .app bundle, not just an Electron helper process.

Examples:
- Xcode Simulator:
  /Applications/Xcode.app/Contents/Developer/Applications/Simulator.app
  bundle id: com.apple.iphonesimulator

- Xcode Instruments:
  /Applications/Xcode.app/Contents/Applications/Instruments.app
  bundle id: com.apple.dt.Instruments

- Parallels VM:
  /Applications/Parallels Desktop.app/Contents/MacOS/Parallels VM.app
  bundle id: com.parallels.vm

These are normal focusable apps/windows, but their bundle path is nested under another outer .app. The behavior to test is: when the foreground app resolves to the nested bundle, Mouser should include both the inner app identity and the outer containing .app identity, so a specific nested profile can win while an outer-app profile can still work as fallback.

farfromrefug:
@chenqianhe thanks will try with ios simulator

TomBadash:
Thanks @chenqianhe! This was integrated into the Logitech MX Master 4 support that just shipped in v3.7.0 (merged via #234), with some hybrid/adjusted merges. You're credited in the README acknowledgments. Closing as integrated, and I really appreciate the contribution. If you spot anything off in how it landed, please let me know.

PR #198, opened 2026-06-02
Handle fractional macOS horizontal scroll deltas

liuxuAP369:
macOS can report Logitech horizontal tilt wheel movement as small fractional deltas. With the default threshold at 1, devices like the M720 require many tilt events before a mapped horizontal-scroll action fires, making desktop/browser mappings appear unresponsive. Lower the default threshold to the existing minimum effective value and allow saved integer thresholds to remain compatible with the float default.

Constraint: Observed M720 macOS delta is about 0.100006 per tilt event

Rejected: Device-specific M720 special case | other Logitech devices can emit the same fractional macOS wheel deltas

Confidence: high

Scope-risk: narrow

Tested: python -m unittest tests.test_config tests.test_engine

Not-tested: Full app GUI regression

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -91,7 +91,7 @@ DEFAULT_CONFIG = {
     "settings": {
         "start_minimized": True,
         "start_at_login": False,
-        "hscroll_threshold": 1,
+        "hscroll_threshold": 0.1,
         "invert_hscroll": False,  # swap horizontal scroll directions
         "invert_vscroll": False,  # swap vertical scroll directions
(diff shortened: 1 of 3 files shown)


TomBadash:
Thanks @liuxuAP369! This was integrated into the Logitech MX Master 4 support that just shipped in v3.7.0 (merged via #234), with some hybrid/adjusted merges. You're credited in the README acknowledgments. Closing as integrated, and I really appreciate the contribution. If you spot anything off in how it landed, please let me know.

PR #201, opened 2026-06-03
chore(build): drop hardcoded macOS signing team default

hughesyadaddy:
## Summary

- Remove the hardcoded Apple team ID fallback from \scripts/build_and_install_macos_app.sh\.
- macOS install builds now require an explicit \MOUSER_SIGN_IDENTITY\ or \MOUSER_TEAM_ID\ (or values from a gitignored \.env.local\).
- Gitignore local QML probe tooling that defaults to machine-specific install paths.

## Why

The install script shipped a developer-specific signing team ID as the default. That is local configuration, not something that belongs in the shared repo.

## Test plan

- [ ] On macOS with \MOUSER_SIGN_IDENTITY\ set: \./scripts/build_and_install_macos_app.sh\ still builds and installs.
- [ ] With neither \MOUSER_SIGN_IDENTITY\ nor \MOUSER_TEAM_ID\ set: script fails fast with a clear error.
- [ ] \.env.local\ remains gitignored (\.env.*\ in \.gitignore\).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ build/
 *.egg
 
 # IDE
-.vscode/
 .idea/
 .cursor/
 *.swp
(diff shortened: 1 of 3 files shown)


hughesyadaddy:
Superseded by #202, which includes the signing-hygiene changes in the same branch. Closing to keep review in one place.

PR #210, opened 2026-06-09
chore(build): enforce env-only code signing with a repo-wide guard test

hughesyadaddy:
## Summary

- Add `.env.local.example` documenting the signing/install environment variables (`MOUSER_SIGN_IDENTITY`, `MOUSER_TEAM_ID`, `MOUSER_INSTALL_DIR`, `MOUSER_PYTHON`), kept tracked via a `.gitignore` negation while `.env*` itself stays ignored.
- Add `tests/test_no_hardcoded_signing_secrets.py`: a repo-wide guard that scans every git-tracked text file and fails if a hardcoded Apple team ID, a literal 40-hex codesigning identity, or a shell/Python fallback default for the signing env vars is ever reintroduced -- and verifies `.env.local` stays gitignored.

## Why

#201 removed a hardcoded signing team default and was closed into #202, but nothing structurally prevents the same class of leak from coming back in a future build-script change. This makes the policy executable: signing configuration can only come from the caller's environment or a gitignored `.env.local`, enforced by CI on every PR.

The guard test builds the previously-leaked team ID at runtime from fragments, so the banned literal never appears in the repo -- including in the test itself.

## Test plan

- [x] `pytest tests/test_no_hardcoded_signing_secrets.py -q` -- 5 passed.
- [x] Full `pytest tests/ -q` on this branch: 500 passed, 1 skipped, 170 subtests passed.
- [x] Verified the guard catches each violation class by temporarily planting a team-ID literal, a `${MOUSER_TEAM_ID:-X}` shell default, and an `environ.get("MOUSER_SIGN_IDENTITY", "X")` default.
- [x] `.env.local.example` is tracked; `.env.local` / `.env` are ignored (`git check-ignore`).

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -1,0 +1,24 @@
+# Copy this file to .env.local and fill in your values.
+# .env.local is gitignored and must NEVER be committed -- it carries
+# machine-specific signing identities and install paths.
+#
+# Build/install scripts that support .env.local load it automatically
+# (values already present in the environment take precedence). For
+# scripts that don't, export these variables in your shell instead.
+
+# -- macOS code signing (set exactly one of the two) -------------------
+# Your Apple Developer Team ID (10 characters). The build resolves the
+# first matching codesigning identity in your keychain.
+#MOUSER_TEAM_ID=
+
+# Or pin an exact codesigning identity (SHA-1 hash form preferred; list
+# yours with: security find-identity -v -p codesigning).
+#MOUSER_SIGN_IDENTITY=
+
+# -- Optional overrides ------------------------------------------------
+# Install destination (default: /Applications on macOS, per-user app
+# directory on Windows).
+#MOUSER_INSTALL_DIR=
+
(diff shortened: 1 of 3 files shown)


thisislvca:

Thanks for putting this together. I like the direction, but I don't think the guard and the docs are fully lined up yet.

The test only catches `${MOUSER_*:-...}` and `os.environ.get(...)` right now, so shapes like `${MOUSER_TEAM_ID-...}`, `${MOUSER_TEAM_ID:=...}`, `os.getenv(...)`, or other wrapper `.get(...)` calls still slip past it.

Also `.env.local.example` is a little ahead of the implementation: `MOUSER_SIGN_IDENTITY` and `MOUSER_PYTHON` are real today, but `MOUSER_TEAM_ID`, `MOUSER_INSTALL_DIR`, and the “scripts load .env.local automatically” story do not look wired up in the current build path yet.

Once those line up I’m happy to re-check.

PR #211, opened 2026-06-09
feat(devices): add Logitech G502 family to the device catalog

hughesyadaddy:
## Summary

- Catalog four G502 variants with product IDs verified against Solaar's device descriptors:
  - **G502 HERO / SE HERO** -- USB `0xC08B`
  - **G502 LIGHTSPEED** -- USB `0xC08D`, Lightspeed receiver WPID `0x407F`
  - **G502 X / X LIGHTSPEED / X PLUS** -- `0xC099` / `0xC098` / WPID `0x409F`
  - **G502 (Proteus Core / Spectrum)** -- `0xC07D` / `0xC332`
- New `G502_BUTTONS` set: middle, back/forward, wheel tilt left/right.
- Non-interactive `g502` placeholder layout (no device art contributed yet) with a user-facing note; the layout invariant test is relaxed to "interactive iff hotspots exist", matching what CONTRIBUTING_DEVICES.md step 3b already documents.

## Why no HID++ buttons or gestures

G-series gaming mice run onboard profiles and do not expose `REPROG_CONTROLS_V4` (0x1B04) -- confirmed in Solaar feature dumps for `C08B` and `C07D` -- so button diversion (gesture / mode_shift / dpi_switch) is impossible. Only the buttons the default onboard profile emits as standard OS events are offered. DPI up/down and the sniper button are consumed onboard and never reach the OS. `ADJUSTABLE_DPI` (0x2201) **is** exposed, so the DPI slider works, with per-sensor clamp ranges (100-25600 HERO/X, 200-12000 Proteus).

## Test plan

- [x] 7 new tests: PID resolution for each variant (incl. receiver WPID), alias resolution (`G502 X PLUS`, `Tunable FPS Gaming Mouse G502`), OS-level-only button gating, placeholder layout fallback, DPI clamping.
- [x] Full `pytest tests/ -q` on this branch: 502 passed, 1 skipped, 200 subtests passed.
- [ ] On-hardware verification welcome from G502 owners -- a Settings -> Debug -> "Copy device info" dump in this PR thread would let us confirm the reprog/feature assumptions per variant.

### DIFF
--- a/core/logi_device_catalog.py
+++ b/core/logi_device_catalog.py
@@ -227,7 +227,7 @@ LOGI_DEVICE_SPECS = (
     {
         "key": "g502_x",
         "display_name": "G502 X",
-        "product_ids": (0xC099, 0xC098, 0x409F),
+        "product_ids": (0xC099, 0xC098, 0xC095, 0x409F, 0x4099),
         "aliases": (
             "G502 X Gaming Mouse",
             "G502 X LIGHTSPEED",
--- a/core/logi_device_catalog.py
+++ b/core/logi_device_catalog.py
@@ -27,6 +27,21 @@ MX_ANYWHERE_SMARTSHIFT_BUTTONS = (
     "mode_shift",
 )
 
+# G502 family (G-series gaming mice). These run onboard profiles and do not
+# expose REPROG_CONTROLS_V4 (0x1B04), so HID++ button diversion -- gesture,
+# mode_shift, dpi_switch -- is unavailable. The buttons below are the ones the
+# firmware emits as standard OS events in its default profile: middle click,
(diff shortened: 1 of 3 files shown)


hughesyadaddy:
Updated this branch so the g502 placeholder layout is manual-selectable in the dropdown. That ensures G502 owners can still choose the correct button set when their device shows up under an unrecognized PID/name (for example through a receiver path). I also ran the full suite on this branch: 503 passed, 1 skipped.

thisislvca:
quick maintainer follow-up: I pushed the missing G502 X Plus ids (`0xC095` / `0x4099`) and added a direct product-id regression test for them. local `python3 -m unittest tests.test_logi_devices` is green here. if CI comes back clean I’m good merging this.

PR #212, opened 2026-06-09
feat(devices): add Logitech G602 to the device catalog

hughesyadaddy:
## Summary

- Catalog the **G602** by its wireless PID `0x402C`. It pairs only through its dedicated receiver (USB `046d:C537`), whose PID is intentionally **not** listed -- Mouser already resolves receiver-paired devices via slot probing + the HID++ device name, same as Unifying/Bolt.
- New `G602_BUTTONS` set: middle, back, forward (OS-level only).
- `g602` placeholder layout with a note explaining that the six side buttons (G4-G9) and top G10/G11 stay with the onboard profile.
- DPI clamp set to the sensor's real 250-2500 range.

## Why OS-level only

Per the Solaar feature dump for this device (WPID `402C`), the G602 speaks HID++ 2.0 but exposes **no REPROG_CONTROLS feature in any version** -- its feature table is battery, `VERTICAL_SCROLLING`, `MOUSE_POINTER`, `ADJUSTABLE_DPI`, and internal/hidden features. Button diversion over HID++ is therefore impossible; only the standard OS button events are mappable. There is no tilt wheel, so no hscroll keys.

## Stacked on #211

This branch is based on `feat/g502-device-support` (#211) because it reuses the placeholder-layout pattern and the relaxed layout invariant test introduced there. Merge #211 first; this PR then reduces to the G602 catalog entry + tests.

## Test plan

- [x] 4 new tests: WPID resolution, reported-name resolution, OS-level-only button set, connected-info/layout/DPI clamping.
- [x] Full `pytest tests/ -q` on this branch: 506 passed, 1 skipped, 205 subtests passed.
- [ ] On-hardware verification welcome from G602 owners -- a Settings -> Debug -> "Copy device info" dump in this PR thread would confirm the receiver/WPID path end to end.

### DIFF
--- a/core/logi_device_catalog.py
+++ b/core/logi_device_catalog.py
@@ -27,6 +27,21 @@ MX_ANYWHERE_SMARTSHIFT_BUTTONS = (
     "mode_shift",
 )
 
+# G502 family (G-series gaming mice). These run onboard profiles and do not
+# expose REPROG_CONTROLS_V4 (0x1B04), so HID++ button diversion -- gesture,
+# mode_shift, dpi_switch -- is unavailable. The buttons below are the ones the
+# firmware emits as standard OS events in its default profile: middle click,
+# back/forward side buttons, and wheel tilt left/right. The DPI up/down and
+# sniper buttons are consumed onboard and never reach the OS. ADJUSTABLE_DPI
+# (0x2201) is exposed, so the DPI slider works.
+G502_BUTTONS = (
+    "middle",
+    "xbutton1",
+    "xbutton2",
+    "hscroll_left",
+    "hscroll_right",
+)
+
 
 def _hotspot(
     button_key: str,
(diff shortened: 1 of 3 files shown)


hughesyadaddy:
Updated this branch so the g602 placeholder layout is now manual-selectable in the dropdown, which keeps a path for owners whose device appears with an unrecognized PID/name. This branch is also synced with the updated #211 base (including the g502 dropdown-selectable fix). Full suite passes here as well: 507 passed, 1 skipped.

hughesyadaddy:
Corrected the button documentation after checking Logitech's official quick-start guide: G4/G5 are the standard Forward/Back OS mouse buttons (already covered by xbutton2/xbutton1), G7-G9 type keyboard keys 1/2/3 from onboard memory, G6 is an internal battery-check with no OS event, and G10/G11 adjust DPI onboard. The mappable button set is unchanged; only the catalog comment and layout note were inaccurate.

thisislvca:
Thanks for updating this. I re-checked the branch, and the catalog/tests themselves look good.

The one thing I'm still not convinced on is the actual runtime path on macOS/Windows. From what I can tell, we only surface the mouse as connected once the HID++ path succeeds and builds connected-device info off the REPROG_V4 path. If the G602 really does not expose REPROG_V4, then making the G602 layout manual-selectable does not fully solve it yet, because that picker only shows after the mouse is already considered connected.

If you've tested this on hardware and the G602 still comes up connected cleanly over the receiver path on current Mouser, can you drop that here? A quick debug dump or confirmation from macOS/Windows would be enough for me to re-check fast. The alias/name cleanup is follow-up, not the thing I'm blocking on.

(btw disregard yesterday's comment, I posted it on the fly while reviewing a bunch of PRs).

DarthMill:
Hi, I have a G602 and I see that there is a need for a dump, I don't have much experience with this software but when I open the app in debug mode from the system tray (windows) and click the copy device info button it prompts that there is no device connected. I have built the app from source on this branch and done the same steps and got the same results. Let me know if there is any thing more I can help with.

Issue #178, opened 2026-05-15
[Bug] Mouser prevents system sleep and screen timeout


mKoonrad:
### Mouser version

v3.6.0

### How are you running Mouser?

Official GitHub release asset

### Operating system

Windows 11

### OS version details

25H2

### Mouse connection type

Bluetooth

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

When Mouser app is running in the background, it prevents Windows 11 from entering sleep mode and prevents screen from turning off according to the system power settings.

### Expected behavior

System should be able to turn off the display and enter sleep mode regardless of whether the app is running in the background.

### Steps to reproduce

1. Open the Moouser application.
2. Minimize it or let it run in the background.
3. Wait for the scheduled system sleep or screen timeout.

### Device info JSON

```json
{
  "device_key": "mx_master_3",
  "display_name": "MX Master 3",
  "product_id": "0xB023",
  "product_name": "Wireless Mouse MX Master 3",
  "transport": "Bluetooth",
  "ui_layout": "mx_master_3",
  "supported_buttons": [
    "middle",
    "gesture",
    "gesture_left",
    "gesture_right",
    "gesture_up",
    "gesture_down",
    "xbutton1",
    "xbutton2",
    "hscroll_left",
    "hscroll_right",
    "mode_shift"
  ],
  "gesture_cids": [
    "0x00C3",
    "0x00D7"
  ],
  "dpi_range": [
    200,
    8000
  ],
  "discovered_features": {
    "REPROG_V4 (0x1B04)": "index 0x09",
    "ADJUSTABLE_DPI (0x2201)": "index 0x0C",
    "SMART_SHIFT (0x2110)": "index 0x0D",
    "BATTERY (0x1000)": "index 0x08"
  },
  "reprog_controls": [
    {
      "index": 0,
      "cid": "0x0050",
      "task": "0x0038",
      "flags": "0x0401",
      "mapped_to": "0x0050",
      "mapping_flags": "0x0000"
    },
    {
      "index": 1,
      "cid": "0x0051",
      "task": "0x0039",
      "flags": "0x0401",
      "mapped_to": "0x0051",
      "mapping_flags": "0x0000"
    },
    {
      "index": 2,
      "cid": "0x0052",
      "task": "0x003A",
      "flags": "0x0531",
      "mapped_to": "0x0052",
      "mapping_flags": "0x0000"
    },
    {
      "index": 3,
[...]

PR #225, opened 2026-07-01
fix: avoid background HID polling resetting system idle timers

ArtisticZhao:
## Summary

This changes Mouser's background HID polling so it no longer keeps resetting the OS idle timer while the app is hidden or the user is inactive.

The issue appears to come from periodic HID++ reads, especially SmartShift polling, being interpreted by the OS as user activity. That prevents display timeout and system sleep even when the mouse and keyboard are untouched.

## Changes

- Gate background battery / SmartShift HID reads behind frontend visibility.
- Skip background HID polling once the system has been idle for a grace period.
- Avoid repeated background polls unless there has been real user activity since the last poll.
- Increase background polling intervals to reduce unnecessary HID traffic.
- Track QML window visibility, including minimized state, and pass it to the engine.
- Add tests covering hidden frontend, idle system state, and repeated polling without user activity.

## Related

Related to #178.

The discussion there identified periodic HID++ reads, especially SmartShift polling, as a likely cause of the OS idle timer being reset while Mouser runs in the background. This PR makes those background reads visibility- and idle-aware, and I manually verified the macOS sleep behavior after building the updated app locally.

### DIFF
--- a/core/engine.py
+++ b/core/engine.py
@@ -4,6 +4,7 @@ current configuration.  Sits between the hook layer and the UI.
 Supports per-application auto-switching of profiles.
 """
 
+import sys
 import threading
 import time
 from core.mouse_hook import MouseHook, MouseEvent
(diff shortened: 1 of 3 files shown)


TomBadash (approved):
Approving. Verified with a before/after idle probe (GetLastInputInfo) on Windows: on master the idle counter resets to 0 every 15s while the machine is untouched (the SmartShift poll), which is what blocks display sleep; on this branch it climbs with no resets while the window is hidden. Engine tests including the three new idle-gating tests pass. Two asks before merge: please rebase onto current master (this branch is on an older base and overlaps `_battery_poll_loop` with #239/#232), and note the intended tradeoff that background battery polling pauses while the window is hidden.

PR #226, opened 2026-07-04
修复 UWP 应用收不到侧键触发的复制粘贴

zeward-lee:
-


TomBadash (approved):
Approving. Adding hardware scan codes to injected key events is the standard fix for UWP/Store apps that ignore vk-only input. Verified copy and paste via a remapped button land correctly in Windows 11 Notepad, and key_simulator tests pass. Low risk. Optional follow-up: a small regression test asserting `wScan` is populated.

Issue #213, opened 2026-06-11
[Feature] Possible to add Scrolling Force setting on Anywhere 3?


Nordhorn:
### Feature description

Logitech options could set the scrolling force, but it seems this setting is missing in mouser.
Possible to add this function?
Thank you!

<img width="1059" height="731" alt="Image" src="https://github.com/user-attachments/assets/1b956819-d289-4ddc-adfb-3f7f8863583d" />

<img width="943" height="635" alt="Image" src="https://github.com/user-attachments/assets/a3438f28-96f2-4e32-af9a-3c1f9650e7fb" />

### Use case / motivation

When you scroll in Ratchet mode, you can adjust the scroll feel of your wheel by moving the Scroll force slider on the Point & Scroll tab. This is a important function.

### Area of the app

Button remapping / actions

### OS relevance

None

### Device info JSON (if relevant)

```json

```

thisislvca:
Scroll force is supported now. Closing this.

PR #232, opened 2026-07-08
Add scroll force support

mizi:
This PR implements https://github.com/TomBadash/Mouser/issues/213

image

## Summary

Adds "scroll force" (ratchet firmness) control for enhanced SmartShift (0x2111) mice.

- New `scroll_force` byte (1-100) in the SmartShift HID++ write/read payload, enhanced (0x2111) devices only - basic (0x2110) devices always get a no-op byte.
- New Scroll Force slider in the Scroll settings page, shown only when the connected device supports it.
- Config bumped to v10 with a `scroll_force` migration.

## Design notes

- `scroll_force` is treated exactly like `smart_shift_threshold`: a concrete default (`50`), always clamped and written on every SmartShift call.
- Both sliders (scroll force and the pre-existing SmartShift threshold slider) now resync to `backend.scrollForce`/`backend.smartShiftThreshold` after a hardware read-back, guarded by `!pressed` so an active drag isn't clobbered. QML slider bindings break permanently on first user interaction (standard QML behavior), so without this the sliders would freeze on whatever was last dragged and never reflect a later device read.

## Also included

- Fix `Main.qml`: `root.forceActiveFocus()` doesn't exist on `ApplicationWindow` (that method belongs to `Item`) — call it on `root.contentItem` instead.
- Fix `Main.qml`: `StandardKey.Close` resolves to multiple platform key bindings (Cmd+W, Cmd+F4, plus a dedicated `Close` key on some keyboards); use `sequences: [...]` instead of `sequence:` so all of them are actually registered instead of silently only the first. Before this change, neither Cmd-W, nor Cmd-F4 would close the window. Only the dedicated `Close` key would, which I didn't even know how to invoke.

## Test plan

- [x] All unit tests pass
- [x] 28 tests in `tests/test_smart_shift.py` cover HID++ write/read, engine persistence/replay, backend properties/slots, and config migration for `scroll_force`
- [x] Manually verified against a real MX Master 4 (enhanced SmartShift) - device connects, reports `scroll_force`, slider reflects the live value

## Translations
Note that I used AI for Chinese translations. Hopefully they look fine, but someone should verify. See the screenshot.

image


TomBadash:
Tested and the feature works: on an enhanced-SmartShift (0x2111) MX Master 3S the Scroll Force slider writes the value and the device echoes it back on read (confirmed at 77), and it persists. Before I approve for merge, could you rebase onto current master? It currently conflicts, and the config v10 bump overlaps with the keyboard PR #237 which also rewrites config.py, so we should sequence those. Happy to approve once it is rebased.

PR #216, opened 2026-06-16
feat(macos): support physical scroll wheel tilt and reduce remapping latency

Archetipo95:
This PR enables physical scroll/tilt wheels (like the Logitech MX Anywhere 2 left/right tilt) to function reliably on macOS even when **Ignore trackpad** is enabled, and adds support for ultra-low horizontal scroll thresholds to eliminate delay when triggering mapped keystrokes.

### Key Changes
1. **Phase-Aware Trackpad Filtering:** In `core/mouse_hook_macos.py`, we inspect `kCGScrollWheelEventScrollPhase` and `kCGScrollWheelEventMomentumPhase`. Trackpads and Magic Mice always generate gesture events with non-zero phases, while traditional mice/tilt-wheels do not (both phases are `0`). We now verify if these phases are non-zero before filtering, letting physical wheel tilt inputs pass through even when `ignore_trackpad` is enabled.
2. **Lower Horizontal Scroll Threshold Limit:** In `core/engine.py`, we lowered the minimum threshold cap to `0.01` (from `0.1`), allowing users to specify a threshold like `0.1` or lower in their configuration. This fires mapped actions instantly on the very first tilt event.
3. **Split joint hscroll UI hotspots:** We split the joint hotspot into `hscroll_left` and `hscroll_right` using the `"mapping"` type instead of `"hscroll"` in `core/logi_device_catalog.py`, allowing them to be mapped separately in the UI.
4. **Validation and test updates:** Updated `core/config.py` type validation to allow floats for numeric default config templates, and updated the mock setups in `tests/test_mouse_hook.py` and `tests/test_device_layouts.py`.

PR #195, opened 2026-06-01
feat: add ping-pong desktop cycling action (macOS only)

winniesi:
## Summary

Adds a new **Cycle Desktops** action that ping-pong cycles through desktops on macOS: `1→2→3→4→3→2→1→2→3…`  This is useful when you want to quickly switch back and forth between desktops without reaching the edge and stopping.

The action appears in the **Navigation** category and can be assigned to any button (mode shift, middle click, side buttons, etc.).

### Desktop detection

Uses two macOS APIs to detect the current state:

1. **`CGSGetActiveSpace`** (private CoreGraphics API) — returns the `id64` of the space on the display where the cursor is located.
2. **`com.apple.spaces`** (user defaults) — contains the per-monitor space configuration, including the ordered list of space IDs for each display.

The implementation parses the `Spaces = (...)` array block for each monitor section using parenthesis-depth matching, and finds the display whose space list contains the current `id64`. This correctly handles multi-monitor setups where each display has its own set of desktops.

### Ping-pong logic

The engine tracks the current direction (`right` or `left`) in `config.json` under `settings.desktop_direction`:

- Moving right until reaching the last desktop → reverse direction
- Moving left until reaching the first desktop → reverse direction
- Only one desktop → skip (no action)

### Platform support

| Platform | `cycle_desktops` action | Desktop detection |
|----------|------------------------|-------------------|
| macOS | ✅ Added | `CGSGetActiveSpace` + `com.apple.spaces` |
| Windows | ❌ Not added | — |
| Linux | ❌ Not added | — |

The macOS `space_right` / `space_left` actions already use `_post_symbolic_hotkey` which triggers the system "Move left/right a space" shortcuts. The `cycle_desktops` action delegates to these existing actions.

## Changes

- **`core/key_simulator.py`**: Add `cycle_desktops` action definition in the macOS `ACTIONS` dictionary (Navigation category). Not added to Windows/Linux.
- **`core/engine.py`**: Add `_get_macos_desktop_info()` static method for desktop detection, and rewrite `_cycle_desktops()` with ping-pong logic.
- **`ui/locale_manager.py`**: Add Chinese translations (zh_CN: 循环切换桌面, zh_TW: 循環切換桌面).

## Testing

1. Open Mouser and connect a Logitech mouse
2. Map any button to **Cycle Desktops** (Navigation category)
3. Create multiple desktops (Mission Control → +)
4. Press the mapped button repeatedly
[...]

PR #184, opened 2026-05-19
Rework swipe-gesture recognition and speed up HID++ device detection

laziukdavid:
## Summary

This PR adds two core reliability/performance reworks, plus a few smaller
improvements, all developed and tested against an MX Master 4 (Logi Bolt
receiver, receiver PID `0xC548`).

### Swipe-gesture recognizer rewrite

The old gesture-button swipe logic summed every movement delta into one running
total, fired when it crossed `gesture_threshold`, then blocked input for a
fixed cooldown. That dropped inputs (the cooldown swallowed the start of the
next swipe), produced false positives (slow drift summed past threshold; the
return stroke between two swipes fired the *opposite* direction), and made
back-to-back repeats impossible.

- New `core/gesture_recognizer.py` — a pure, thread-safe `GestureRecognizer`
  that segments motion into *strokes* instead of integrating one global sum.
  A stroke commits a swipe on threshold travel along one axis within a commit
  window and within an off-axis ratio; a peak detector fires every fresh flick
  while return strokes only re-arm it; a pause clears the lock. No cooldown.
- Shared by the macOS, Windows and Linux hooks — removes the triplicated
  `_accumulate_gesture_delta` / `_detect_gesture_event` code.
- Click-jolt rejection (minimum movement-report count) and quick-flick capture
  fix phantom swipes on a tap and missed swipes during a fast flick.
- Config schema **v9 → v10**: retires `gesture_deadzone` / `gesture_timeout_ms`
  / `gesture_cooldown_ms`; adds `gesture_commit_window_ms`, `gesture_settle_ms`,
  `gesture_cross_ratio`. Obsolete keys in existing configs are left ignored.
- The raw `gesture_threshold` px slider is replaced with a discrete 5-level
  **Sensitivity** dropdown (stored key/units unchanged).

### HID++ detection speed

Cold-start connection on a setup with several Logitech receivers was ~46s,
almost entirely spent probing the wrong devices before reaching the mouse.

- Receiver-aware device-index scan order
- Persistent device cache (`device_cache.json`) so a relaunch / reconnect goes
  straight to the last-good device
- Dead-slot early-out — stop probing a slot once its control-count read fails
- 600ms discovery probe timeout applied across the **whole** discovery phase
  (operational requests keep the 2s timeout)
- Candidate dedup — a multi-interface device (e.g. a webcam exposing vendor
  usage pages) is opened once, not once per interface

## Additional changes

[...]

Issue #119, opened 2026-04-21
[Device] MX Master 4 haptic button


ChrisWhitling:
### Device name and model

Logitech MX Master 4

### Product ID (if known)

0xC548

### How is the device connected?

USB receiver (Bolt / Unifying)

### Does Mouser detect the device at all?

Yes, it shows the correct model name

### Device info JSON

```json
{
  "device_key": "mx_master_4",
  "display_name": "MX Master 4",
  "product_id": "0xC548",
  "product_name": "MX Master 4",
  "transport": "hidapi",
  "ui_layout": "mx_master_4",
  "supported_buttons": [
    "middle",
    "gesture",
    "gesture_left",
    "gesture_right",
    "gesture_up",
    "gesture_down",
    "xbutton1",
    "xbutton2",
    "hscroll_left",
    "hscroll_right",
    "mode_shift"
  ],
  "gesture_cids": [
    "0x00C3",
    "0x00D7"
  ],
  "dpi_range": [
    200,
    8000
  ],
  "discovered_features": {
    "REPROG_V4 (0x1B04)": "index 0x0D",
    "ADJUSTABLE_DPI (0x2201)": "index 0x14",
    "SMART_SHIFT_ENHANCED (0x2111)": "index 0x11",
    "BATTERY (0x1004)": "index 0x09"
  },
  "reprog_controls": [
    {
      "index": 0,
      "cid": "0x0050",
      "task": "0x0038",
      "flags": "0x0401",
      "mapped_to": "0x0050",
      "mapping_flags": "0x0000"
    },
    {
      "index": 1,
      "cid": "0x0051",
      "task": "0x0039",
      "flags": "0x0401",
      "mapped_to": "0x0051",
      "mapping_flags": "0x0000"
    },
    {
      "index": 2,
      "cid": "0x0052",
      "task": "0x003A",
      "flags": "0x0531",
      "mapped_to": "0x0052",
      "mapping_flags": "0x0000"
    },
    {
      "index": 3,
      "cid": "0x0053",
      "task": "0x003C",
      "flags": "0x0531",
      "mapped_to": "0x0053",
      "mapping_flags": "0x0000"
    },
    {
      "index": 4,
      "cid": "0x0056",
      "task": "0x003E",
      "flags": "0x0531",
      "mapped_to": "0x0056",
      "mapping_flags": "0x0000"
    },
    {
      "index": 5,
      "cid": "0x00C3",
      "task": "0x009C",
      "flags": "0x0531",
[...]

PR #234, opened 2026-07-09
feat: MX Master 4 support (Actions Ring, haptics, gestures, dual-CID HID++)

davidnoyes:
This branch adds MX Master 4 support and related improvements, building on a number of community PRs (with some hybrid/adjusted merges). Ref #119.

## Summary
- MX Master 4 firmware-first HID++ runtime, dual-CID routing (Sense Panel + thumb Gesture button), and fast Bluetooth device detection
- Actions Ring radial overlay with a glassmorphism UI and animations
- HID++ haptic feedback: global toggle, per-action and per-button pickers, and dedup
- Stroke-aware gesture recognizer; each gesture-capable button has its own independent movement-swipe set, active when its tap action is "Do Nothing"
- Gesture-button force sensitivity (HID++ 0x19C0)
- Config keys track physical buttons, with device-appropriate event families so MX Master 3/3S behaviour is unchanged
- MX Master 3/3S and MX Master 4 overlay hotspot layouts
- New first-install defaults tuned for the MX Master 4

## Credit
This work integrates and adapts the following PRs. Thanks to their authors:

- @hughesyadaddy - #171 (firmware-first HID++ runtime), #180 (DPI persistence), #181 (Backend memoization), #185 (OS event pass-through)
- @laziukdavid - #184 (swipe-gesture rework and faster HID++ detection)
- @FunJim - #191 (Shift+wheel horizontal scroll), #193 (symlink-safe config save)
- @chenqianhe - #196 (nested macOS app identity matching)
- @liuxuAP369 - #198 (fractional horizontal scroll deltas)
- @winniesi - #195 (ping-pong desktop cycling)
- @Archetipo95 - #216 (scroll-wheel tilt and reduced latency, partial)

## ⚠️ Note for pre-release testers

The config schema was consolidated — the unreleased dev versions (10–20) are squashed into a single `v9 → v10` migration. If you ran an earlier build of this branch, your `config.json` may be stamped at an old dev version (11–20) and **won't auto-migrate**; delete it to pick up clean defaults:

- macOS: `~/Library/Application Support/Mouser/config.json`
- Windows: `%APPDATA%\Mouser\config.json`
- Linux: `~/.config/Mouser/config.json`

Public release users (v3.6.0, config v9) are unaffected — they upgrade normally.


core/engine.py
```diff
@@ -613,6 +1102,13 @@ def _on_connection_change(self, connected):
         hid_features_changed = hid_features_ready != self._last_hid_features_ready
         if connection_changed:
             self._last_connection_state = connected
+            if connected:
+                # Re-wire hooks now that the device (and its
+                # gesture_via_sense_panel / supported_buttons) is known, so the
+                # per-device gesture recognizers and rawXY hand-off are applied.
+                with self._lock:
+                    self.hook.reset_bindings()
+                    self._setup_hooks()
```
chenqianhe: Avoid waiting for the HID listener from its own connection callback

`_on_connection_change()` is called synchronously on the `HidGestureListener` thread:

1. `HidGestureListener._main_loop()` calls `self._on_connect()` immediately after `_try_connect()` succeeds.
2. `BaseMouseHook._on_hid_connect()` calls `_set_device_connected(True)`.
3. `_set_device_connected()` invokes the Engine's `_on_connection_change(True)` callback.
4. This newly added branch calls `_setup_hooks()`.
5. `_setup_hooks()` calls `_apply_wheel_invert_setting()`.
6. `_apply_wheel_invert_setting()` calls `request_wheel_native_invert()`, which waits up to three seconds for `_wheel_divert_event`.

The problem is that `_wheel_divert_event` can only be completed when the same listener thread returns to `_main_loop()` and processes `_pending_wheel_divert` through `_apply_pending_native_wheel_invert()`. Because the listener is still blocked inside this callback, it cannot process the pending request. The request therefore times out every time this path is taken.

The timeout also clears `_pending_wheel_divert`, so returning to the listener loop does not apply the requested state afterward. On the initial connection of a wheel-capable device, this leaves `wheel_native_invert_active` false and forces the engine onto the OS fallback path. It also adds an avoidable three-second stall to connection handling.

Could we keep the callback-time binding refresh, but defer the blocking wheel write until after the listener callback returns? For example, `_setup_hooks()` could optionally skip `_apply_wheel_invert_setting()`, then a worker thread could call it after the bindings have been refreshed. Alternatively, the listener could expose a non-blocking queue API for calls originating on its own thread.

A regression test should model the real threading relationship: the connect callback runs on the queue-owning listener thread, and completion of the queued write requires that thread to return to its loop.

core/config.py
```diff
@@ -326,10 +541,105 @@ def _migrate(cfg):
...
+        for prof in cfg.get("profiles", {}).values():
+            m = prof.get("mappings", {})
+            if old_mode == "ring":
+                m["actions_ring"] = "activate_actions_ring"
+            elif old_mode == "disabled":
+                m["actions_ring"] = "none"
+            # "simple": keep existing mapping (already a real action ID)
+        cfg["version"] = 19
+
+    if version < 20:
```
chenqianhe: The v20 migration now applies the Actions Ring mapping to the wrong physical control

This migration was written while the configuration keys had the temporary, opposite meaning used by earlier commits in this PR. The final model now explicitly defines the keys by physical button:

- `gesture` = the physical thumb Gesture button, emitting `gesture_*`
- `actions_ring` = the MX Master 4 Sense Panel, emitting `sense_*`

However, this migration still moves `activate_actions_ring` from `actions_ring` to `gesture`.

This produces a concrete upgrade inconsistency. Starting from the actual base branch's v9 configuration, where `gesture` is normally `"none"`, the migration chain currently produces:

    gesture = "activate_actions_ring"
    actions_ring = "none"

A fresh v20 configuration produces the opposite intended setup:

    gesture = "app_expose"
    actions_ring = "activate_actions_ring"

As a result, most existing users upgrading from the current release get the Actions Ring on the thumb Gesture button, while fresh installs get it on the MX Master 4 Sense Panel. This also contradicts the final physical-button documentation and the PR description.

The v19 migration already assigns `activate_actions_ring` to `actions_ring`, which is the correct final key for the Sense Panel. For users upgrading from the real v9 base, the v20 move should therefore not happen. If configurations produced by intermediate versions of this PR also need to be supported, it would be safer to bump the schema again and add a narrowly targeted migration rather than continuing to reinterpret version 20.

Please add an upgrade test using a copy of the base branch's actual v9 default configuration. The test should verify both:

1. Existing custom Gesture-button mappings remain attached to the physical thumb button.
2. The resulting default Actions Ring assignment matches a fresh configuration and targets the Sense Panel.

chenqianhe: Migrating this representative v9 profile:

    {
        "version": 9,
        "profiles": {
            "default": {
                "mappings": {
                    "gesture": "none"
                }
            }
        },
        "settings": {}
    }

currently yields:

    {
        "gesture": "activate_actions_ring",
        "actions_ring": "none"
    }

while `DEFAULT_CONFIG` yields:

    {
        "gesture": "app_expose",
        "actions_ring": "activate_actions_ring"
    }

core/config.py
```diff
@@ -24,92 +24,193 @@
...
+                "hscroll_right": "none",            # pass-through
                 "mode_shift": "switch_scroll_mode",
+                # Sense Panel ("Actions Ring", MX4) — activates the ring.
+                "actions_ring": "activate_actions_ring",
+                "actions_ring_left": "none",
+                "actions_ring_right": "none",
+                "actions_ring_up": "none",
+                "actions_ring_down": "none",
+                "thumb_button": "none",
+                "actions_ring_slots": [
```
chenqianhe: These new defaults contain macOS-only action IDs on every platform

`DEFAULT_CONFIG` is shared by Windows, macOS, and Linux, but the new defaults reference action IDs that only exist in the macOS `ACTIONS` table:

- The default `gesture` mapping is `app_expose`.
- The Actions Ring slots contain `mission_control`, `show_desktop`, and `launchpad`.
- Of the four default ring slots, only `play_pause` exists on all supported platforms.

On Windows, `execute_action()` looks the ID up in the Windows `ACTIONS` dictionary and returns without executing anything when it is missing. Linux follows the same pattern. Therefore, on a fresh Windows or Linux installation:

- The MX Master 4 Gesture button is mapped to `app_expose`, but pressing it does nothing.
- Three of the four default Actions Ring sectors do nothing.
- Only the `play_pause` sector works.

This is especially confusing because the UI still displays the raw unknown IDs as labels through `_resolve_ring_label()`, so the options can appear present even though the platform cannot execute them.

Could these defaults use platform-neutral action IDs, with each platform mapping those IDs to the appropriate native behavior? Another option would be to construct the default mappings per platform. Either way, it would be useful to add an invariant test asserting that every non-custom action referenced by `DEFAULT_CONFIG`, including every Actions Ring slot, exists in the current platform's `ACTIONS` table.

chenqianhe (commented):
Thank you for the shout-out and for integrating my app-detection fix! I really appreciate the care you’ve taken to credit all the original contributors. Bringing so many improvements together into one cohesive MX Master 4 implementation is a huge effort!

core/engine.py
davidnoyes: Confirmed and fixed — thanks, this was exactly right. `_on_connection_change` runs on the HID listener thread, so calling `request_wheel_native_invert` inline could never complete: the listener was blocked in its own callback and couldn't service the queued write, timing out (~3s) every connect and clearing the pending state, which forced the OS fallback.

Fix (`5c64fa5`): `_setup_hooks` now takes a `defer_wheel_invert` flag. The connect path refreshes bindings synchronously but runs the blocking wheel write on a short-lived worker thread (only when the HID runtime supports the write), so the listener returns to its loop and completes the request. Added a regression test that models the threading relationship — it asserts the write does **not** run on the connect callback thread and is applied via the deferred worker.

core/config.py
davidnoyes: Confirmed — the migration did land the ring on the thumb button for upgraders while fresh installs got it on the Sense Panel. Good catch.

Fix (`1dbf1d1`): rather than reinterpret v20, I squashed the entire unreleased pre-release chain (schema 10–20) into a single `v9 → v10` migration and dropped the backwards move. Ring activation now lands on `actions_ring` (Sense Panel) for upgraders and fresh installs alike; the thumb `gesture` key is left untouched so custom mappings (and thumb swipes) are preserved.

One deliberate deviation from your suggestion: you proposed bumping the schema again with a narrowly-targeted migration to also handle configs from intermediate builds of this PR. Since nothing above config v9 was ever released (latest release v3.6.0 ships v9), I opted to squash it away instead of layering another migration — pre-release testers on an intermediate build are asked to reset their config. Added an end-to-end test that upgrades the verbatim v3.6.0 (v9) release config and asserts (1) a custom Gesture-button mapping stays on the thumb, and (2) the default ring assignment matches a fresh install on the Sense Panel.

core/config.py
davidnoyes: Confirmed — `app_expose`, `mission_control`, `show_desktop`, and `launchpad` exist only in the macOS `ACTIONS` table, so fresh Windows/Linux installs had a dead gesture button and three dead ring sectors.

Fix (`1dbf1d1`): the platform-varying defaults are now built per OS — macOS keeps its native actions; Windows/Linux get `task_view`, `win_d`, and `screenshot_region_clip` (all present in their tables, and aligned with the app's own convention where `win_d`/`task_view` are hidden on macOS). Also added the missing overlay labels for the new defaults so the ring never renders a raw id. Added the invariant test you suggested — it asserts every non-custom default action, including every ring slot, exists in the target platform's `ACTIONS` table (checked for all three platforms), plus a check that each default ring slot has a short label.

TomBadash:
Hey @davidnoyes - first off, thank you for taking the [`test-mxm4`](https://github.com/TomBadash/Mouser/tree/test-mxm4) branch and turning it into such great MX Master 4 support. It came together really well, and I know it was a real team effort with everyone who chipped in, as you noted in #119. Genuinely appreciate it.

I pulled the branch down, tested it on a real MX Master 4, and pushed a few additions and fixes on top:

**Additions**
- Haptic feedback as the cursor lands on each Actions Ring slot. It uses an immediate fire-and-forget HID write so there's no lag, and it plays at the configured global haptic level. New toggle for it on the Haptic page.
- Global vs per-app ring actions: a "Same ring for every app" toggle (on by default) with a profile picker for per-app mode. Switching per-app back to global uses only the global set, and per-app data is preserved.

**Fixes**
- The ring-slot "Custom Shortcut" option was saving the literal `__custom__` placeholder instead of opening the key-capture dialog, so those slots did nothing. It now captures keys and stores `custom:<combo>` like the button pickers.
- `angle_to_sector` could throw `ZeroDivisionError` when a profile had the ring enabled but no slots (e.g. a config stamped at an older dev version that skipped the slot backfill). Added a guard so an empty ring can't crash.

All tested end to end on the device, and the migration seeds the global ring from the existing Default profile so nothing is lost on upgrade.

Let me know if you'd like to add or adjust anything before I merge this into `master` and cut a new version. Thanks again for all the work here.

davidnoyes:
Frankly, if you’re happy, I’m happy.
Thank you

PR #237, opened 2026-07-14
feat: Keyboards support — multi-device multiplexing, Craft Crown, MX Keys, generic keyboards

AlessandroTischer:
## Multi-device multiplexing

- The HID++ listener now binds **all** HID++ devices instead of the first one that accepts a gesture divert: every device paired to a USB receiver (slots 1–6 on one handle) plus each direct Bluetooth device (own handle, devIdx `0xFF`).
- Per-device **sessions**: each bound device keeps its own feature indexes, divert set, gesture CIDs, battery state, and device type. Incoming reports are routed to the owning session by handle + device index; the main loop polls every distinct handle, splitting the read budget across them.
- **Hot-plug**: receiver notifications (`0x41` arrival / `0x40` departure) rebind devices that power on/off after the initial scan, without restarting the listener.
- **Per-device battery** reporting (`read_all_batteries`), keyed by device type, shown per page in the UI.
- Divert gating (mode shift, DPI switch) now considers *all* connected devices, so a keyboard being the "primary" device no longer drops the mouse's diverts — and vice versa.
- A mouse and a keyboard sharing a single Unifying receiver work simultaneously.

## Logitech Craft — Crown dial (HID++ feature 0x4600)

- Crown events are diverted and decoded natively: **rotate left/right**, **click (tap)**, **capacitive touch**, and a **click + rotate** layer — six independently remappable controls (defaults: rotate = volume, click = play/pause, click+rotate = track skip; touch is opt-in).
- Touch/click disambiguation window so resting a finger on the Crown doesn't fire the click action.
- Reliable Crown delivery requires enabling the receiver's HID++ notification register (`wireless | software_present`) — done once per connection.
- **Ratchet ↔ smooth rotation feel**, stored **per application profile** and re-applied automatically on profile switch; also exposed as a mappable **Toggle Crown Ratchet/Smooth** action. Craft-only: the UI toggle and the action are only offered while a Crown-equipped keyboard is connected.
- Catalog entry matched by HID++ device name (the Unifying receiver hides the real PID).

## MX Keys and generic keyboard strategy

- A device-agnostic table of standard Logitech **top-row keys** (brightness, backlight, media, calculator, screen capture, context menu, screen lock, …) with their `REPROG_CONTROLS_V4` control IDs. Any keyboard exposes exactly the subset its firmware advertises.
- Keys are diverted **on demand**: a key stays fully native until the user remaps it, at which point only that control is diverted.
- **Auto-classification** of unknown HID++ devices as mouse or keyboard using exclusive features (e.g. ADJUSTABLE_DPI vs BACKLIGHT2/CROWN), REPROG control flags, and product name fallback — unknown keyboards get a generic keyboard card with their real remappable keys.
- MX Keys control set verified on hardware (Bluetooth PID 0xB35B), including its Launchpad/App-Switch CID variant.

## Keyboard backlight (HID++ feature 0x1982)

- **Backlight follows the system theme** (opt-in): on in dark mode, off in light mode. These keyboards expose on/off only, not software brightness.
- While the backlight is auto-disabled, the keyboard's backlight-up key is temporarily diverted so a single press re-enables it (then returns to native).

## UI

- **Separate Mouse and Keyboard pages** in the navigation; both are instances of one device page component filtered by device class, each rendering its own device (multiplexer-aware).
- **Interactive photo layouts** for Craft and MX Keys: clickable top-row keys and Crown dial over device photos, in the same style as the mouse overlays.
- **Crown panel** for assigning all six Crown sub-actions and the rotation feel.
- Hidden **layout calibration tool** (Settings → Developer) that overlays a measuring grid on device photos to derive hotspot coordinates — works without a connected device.
- **Profile copy**: duplicate an existing profile's mappings (including crown feel) into a new one.
- **Open Application…** action: bind any button, key, or Crown sub-action to launch an executable picked from a file dialog.
- Per-device action filtering: DPI cycling hidden on keyboards, crown actions hidden without a Crown, action categories computed per connected device.
- Settings page reorganization (renamed from Scroll page) and additional localized strings (EN / zh-CN / zh-TW) covering all new UI.

## Internals

- **Non-mouse-centric refactor**: `mouse_hook_*` → `device_hook_*`, `MouseHook` → `DeviceHook`, `MouseEvent` → `DeviceEvent`, since hooks now serve keyboards too. Public behavior is unchanged.
- Config schema: keyboard support lands as a single migration step (v10 → v11) mirroring the v9 → v10 MX Master 4 squash; existing mappings are preserved (pre-release fork configs should be reset, as with the MX4 pre-release chain).
- `tools/craft_probe.py`: safe, read-only probe utility that dumps HID++ features, REPROG controls, and key CIDs of a connected keyboard — useful for adding new keyboard entries.
- Tests for the multiplexer (session routing, per-device flags, keyboard binding without gesture CIDs), crown decode, classification, migrations, backlight gating, and keyboard layouts.

## Integration with current master

The fork is merged with v3.7.0: the MX Master 4 runtime (Actions Ring, haptics, force sensing, native wheel invert, warm-start device cache, stroke-aware gesture recognizer) is preserved and extended to be **session-aware** so it coexists with the multiplexer (per-device haptic/wheel/force state, session-aware apply paths).

---


AlessandroTischer:
@TomBadash let me know if there is anything I can do to help this being merged easily 😉

Issue #133, opened 2026-04-29
[Feature]Add a feature that displays the mouse battery level in the taskbar


popuping86:
### Feature description

Add a feature that displays the mouse battery level in the taskbar.

### Use case / motivation

It allows you to see the mouse’s remaining battery level at a glance, avoiding unexpected shutdowns due to low power.

### Area of the app

Other

### OS relevance

None

### Device info JSON (if relevant)

```json

```

Nick-b404:
to add to this, as a first step, it would be nice if Mouser could tell if the mouse is actively charging. Just plugged my MX master 3s in and the battery status isn't responsive, so doesn't show that it is charging.
Maybe there's a way to get more granular battery levels too? But I wouldn't be surprised if that's a proprietary Logitech protocol they use to do that on Logitech Options

PR #239, opened 2026-07-14
feat: show charging indicator on battery badge

davidnoyes:
## What

Adds a charging (lightning-bolt) indicator to the mouse battery badge. When a wireless mouse is plugged in to charge, the battery glyph swaps to a bolt variant so you can see it's charging at a glance instead of waiting for the percentage to climb.

## How

The HID++ battery response already carries a charge-state byte next to the percentage — it was being discarded. This change:

- Parses the charge-state byte from both the Unified Battery (`0x1004`) and Battery Status (`0x1000`) features (`charging` = recharging / near-full / full-on-cable / slow-recharge).
- Handles the device's **unsolicited battery-status broadcast** in the HID listener and pushes it straight to the UI. **This path is independent of the battery poll** — it is event-driven by the device, so the indicator appears the instant USB is plugged in (or is unplugged) without any change to the polling cadence.
- Plumbs a `charging` flag through engine → backend → a new `batteryCharging` QML property, and swaps the badge icon (new `battery-charging.svg`).
- The existing 300s poll path also carries the charging flag, purely as a slow, self-correcting fallback if a device ever misses a broadcast. The poll interval is unchanged.

## Testing

- Tested on **MX Master 3S** and **MX Master 4**: bolt appears instantly on plug-in and clears on unplug.
- `pytest` green (added `set_battery_notify_callback` to the affected hook test doubles).

## Notes

Related to #133 (battery visibility). This adds a charging indicator to the existing in-app battery badge; it does not add the taskbar/menu-bar battery display that #133 specifically requests.


TomBadash (approved):
Approving. Tested on MX Master 3S: the charging bolt appears within a second of plugging in USB and clears on unplug, so the unsolicited-broadcast path works. Automated suite is green. Merge-order note: this, #225, and #232 all touch `_battery_poll_loop`, so whichever lands after the first will need a rebase. Thanks for the clean implementation and for testing on 3S and MX4.

PR #240, opened 2026-07-14
fix(macos): close IOHIDManager to stop reconnect memory and port leak (#238)

### LINKED ISSUE #238: [macOS] Repeated HID reconnects leak IOHIDManager objects and Mach ports, growing footprint to 3.3 GB
## Summary

Mouser appears to leak `IOHIDManager` instances, Mach ports, and substantial heap memory during repeated HID discovery/reconnection attempts on macOS.

After approximately 24 hours, the Mouser process reached:

- 3.3-3.4 GB physical footprint
- 3.2 GB malloc heap
- 11.2 million live allocations
- 33,662 live `IOHIDManager` objects
- approximately 67,862 Mach ports
- 3.3 GB of writable memory swapped out

This is not merely a large virtual-address reservation. Activity Monitor reports the process physical footprint, and `vmmap`, `heap`, and `leaks` confirm retained allocations and leaked objects.

## Environment

- Mouser version: `3.6.0`
- Build commit: `34d93f70b2a84e425698e0d3d748cae2c5d18911`
- Architecture: Apple Silicon / ARM64
- macOS: `26.4.1 (25E253)`
- Device: Logitech MX Master
- Connection: Bluetooth Low Energy
- HID backend: `auto`
- Process uptime at capture: approximately 24 hours
- Mouser PID during capture: `41115`

## Observed behavior

Activity Monitor showed Mouser using approximately 3.30 GB. The process details window showed only about 31.8 MB of resident memory, which initially made this look like a virtual-memory accounting issue. Further inspection showed that most of the process footprint had instead been compressed or swapped out.

### `vmmap -summary`

```text
Physical footprint:         3.3G
Physical footprint (peak):  3.4G

Writable regions:
Total=7.7G
written=3.3G
resident=13.6M
swapped_out=3.3G

MALLOC ZONE TOTAL:
SIZE       7.5G
COUNT      11201047
ALLOCATED  3.2G
```

The low RSS therefore does not mean the memory was released. Approximately 3.3 GB of dirty process memory had been moved to swap.

## Heap analysis

Running:

```bash
heap -s -H 41115
```

produced:

```text
All zones: 11203439 nodes (3.2G)

COUNT       BYTES    CLASS
179550      2.1G     NSConcreteMutableData (Bytes Storage)
3104593     434.8M   non-object
1512348     161.5M   HIDEvent
504170      69.2M    CGSEventAppendix
112263      32.5M    NSXPCConnection._connection
258374      31.5M    dispatch_queue_t (serial)
504170      15.4M    CGEvent
33662       8.9M     IOHIDManager
```

The `IOHIDManager` count is especially suspicious. A normal long-running utility should not retain tens of thousands of HID managers.

## Leak analysis

Running:

```bash
leaks --nostacks --groupByType 41115
```

reported:

```text
Process 41115: 11206177 nodes malloced for 3326632 KB
Process 41115: 3775426 leaks for 240439904 total leaked bytes

3775426 (229M) << TOTAL >>

3606908 (221M) ROOT LEAK: GPProcessMonitor
```

The leaked `GPProcessMonitor` tree contained large numbers of:

```text
NSXPCConnection
NSXPCInterface
dispatch_queue_t
dispatch_group_t
NSLock
GPProcessInfo
GPAutoEDRParameters
NSMutableDictionary
```

There were also over 1.5 million `HIDEvent` objects and approximately 500,000 `CGEvent` / `CGSEventAppendix` objects still present.

This resembles the object signature described in #125, even though the v3.6.0 build already contains the autorelease-pool fix from that PR. The fix may be incomplete or ineffective with the packaged runtime on macOS 26.4.1.

## Mach port growth

`top` reported:

```text
PID    COMMAND  MEM     #PORTS
41115  Mouser   3389M   67862
```

The port count is close to two ports for every retained `IOHIDManager`, suggesting that the managers continue to hold IOKit/Mach resources.

A later 30-second sample was stable at approximately 3.39 GB and 67,862 ports because the latest reconnect storm had stopped. The accumulated memory and ports were not reclaimed.

## Correlation with HID reconnect logs

The six retained rotating log files covered roughly the most recent nine hours of reconnect activity. They contained:

```text
2608   "Backend preference: auto"
20784  "Opened PID=... via iokit..."
7788   "Can't open PID=... via iokit..."
145415 "IOHIDDeviceSetReport failed..."
```

That represents at least 31,180 IOKit manager creation/open paths in the retained logs alone. Older records had already been removed by log rotation. This closely matches the 33,662 `IOHIDManager` objects found in the live heap.

Typical repeated output looked like:

```text
[HidGesture] Backend preference: auto
[HidGesture] Candidate HID interfaces: 8
[HidGesture] Opened PID=0xB017 via iokit-exact
[HidGesture] request tx failed ... IOHIDDeviceSetReport failed
[HidGesture] Opened candidate but REPROG_V4 was not found ...
```

The probing also repeatedly attempted unrelated USB keyboard and mouse interfaces.

## Suspected code paths

In `core/hid_gesture.py`, `IOHIDManagerCreate` and `IOHIDManagerOpen` are declared and used, but `IOHIDManagerClose` is not declared or called.

### Enumeration cleanup

`enumerate_infos()` opens a manager, then only performs:

```python
if manager:
    _cf.CFRelease(manager)
```

It does not call:

```python
_iokit.IOHIDManagerClose(manager, 0)
```

### Device cleanup

`_MacNativeHidDevice.close()` closes the selected `IOHIDDevice`, but manager cleanup is also limited to:

```python
if self._manager:
    _cf.CFRelease(self._manager)
    self._manager = None
```

Apple documents that [`IOHIDManagerClose`](https://developer.apple.com/documentation/iokit/1438405-iohidmanagerclose) closes the manager and all currently enumerated devices.

### Failed-open cleanup

There is another likely leak when `d.open()` raises:

```python
d = _MacNativeHidDevice(...)
d.open()
```

The exception handler clears `self._dev`, but does not call `d.close()`. Because `_MacNativeHidDevice` stores raw Core Foundation pointers and has no destructor that releases them, partially initialized values such as the manager, matching dictionary, and matching references can remain unreleased.

The same missing `IOHIDManagerClose` behavior appears to remain in the current [`core/hid_gesture.py`](https://github.com/TomBadash/Mouser/blob/master/core/hid_gesture.py).

## Steps to reproduce

1. Run Mouser on macOS with the HID backend set to `auto`.
2. Connect an MX Master over Bluetooth.
3. Trigger repeated discovery failures, for example by letting the device disconnect, sleep, or repeatedly fail HID++ feature probing.
4. Confirm that the logs repeatedly show:
   - `Backend preference: auto`
   - `Candidate HID interfaces`
   - `Opened PID ... via iokit`
   - `IOHIDDeviceSetReport failed`
   - `REPROG_V4 was not found`
5. Leave Mouser running for several hours.
6. Inspect the process:

```bash
PID="$(pgrep -x Mouser)"

top -l 1 -pid "$PID" \
  -stats pid,command,cpu,mem,ports,threads,time

vmmap -summary "$PID"
heap -s -H "$PID"
leaks --nostacks --groupByType "$PID"
```

7. Observe that the `IOHIDManager` count, Mach port count, and physical footprint increase after repeated probe/reconnect cycles and are not reclaimed afterward.

## Expected behavior

- HID managers and their Mach resources should be closed after every enumeration or failed connection attempt.
- Reconnect attempts should not cause cumulative heap or port growth.
- Unsupported interfaces should not be probed indefinitely.
- A long-running background instance should retain a bounded amount of memory.

## Suggested fix

1. Declare `IOHIDManagerClose`:

```python
_iokit.IOHIDManagerClose.argtypes = [c_void_p, c_int]
_iokit.IOHIDManagerClose.restype = c_int
```

2. For every successfully opened manager, call:

```python
_iokit.IOHIDManagerClose(manager, 0)
_cf.CFRelease(manager)
```

3. Make `_MacNativeHidDevice.open()` exception-safe:

```python
def open(self):
    try:
        # Current initialization...
    except Exception:
        self.close()
        raise
```

4. Ensure `close()` is idempotent and cleans up:
   - run-loop scheduling
   - input callbacks
   - `IOHIDDevice`
   - `IOHIDManager`
   - matching dictionaries
   - matching keys and values
   - partially initialized state
5. Add exponential backoff or a retry limit for repeated incompatible-interface probes.
6. Avoid probing unrelated USB keyboards/mice after an interface is known not to expose `REPROG_CONTROLS_V4`.
7. Revalidate the autorelease-pool fix from #125 on macOS 26.4.1, since the same `HIDEvent`, `CGEvent`, and `GPProcessMonitor` signature remains present.

## Suggested regression test

A macOS-specific stress test could force at least 100 failed discovery/reconnect cycles and verify that:

- `IOHIDManager` object count remains bounded.
- Mach port count returns close to baseline.
- Physical footprint does not increase linearly.
- Failed `open()` paths release every Core Foundation object.
- Repeated `NSWorkspace` polling and `CGEventTap` callbacks do not accumulate `GPProcessMonitor` or event objects.

TomBadash:
## Summary

Fixes #238. On macOS the IOKit HID backend created `IOHIDManager`
instances with `IOHIDManagerOpen` but only ever `CFRelease`d them,
never calling `IOHIDManagerClose`. Close is what frees the IOKit
user-client Mach ports and the enclosed-device opens, so every
enumeration and every open attempt leaked one manager plus ~2 Mach
ports. During repeated BLE discovery/reconnect cycles this grew the
process to ~3.3 GB with ~33k live `IOHIDManager` objects and ~67k Mach
ports, exactly as @maxhis documented with `vmmap`/`heap`/`leaks`.

## Changes

**Block A: stop the leak**
- Declare `IOHIDManagerClose` and add a `_close_manager()` helper that closes before releasing.
- Route `enumerate_infos()` and `_MacNativeHidDevice.close()` through it.
- Make `open()` exception-safe: on any partial failure it unwinds through the idempotent `close()`, so a failed open can no longer leak the manager, matching dict, CF refs, or the retained device (previously the failed-open path only cleared `self._dev`, and the class had no destructor).
- Add a `__del__` backstop.

**Block B: throttle the reconnect/probe storm that amplified the leak**
- Exponential connect backoff (1.5s to 30s, reset on success) via an interruptible sleep helper, replacing the flat 1.5s retry.
- Short per-interface "REPROG_V4 absent" cooldown so a stable set of incompatible interfaces is not re-probed every pass, bypassed when the device is the sole candidate so a just-woken mouse still reconnects promptly.

## Testing

- `tests/test_hid_gesture.py`: 49 passed (44 existing + 5 new regression tests for the throttle logic).
- The IOKit `IOHIDManagerClose` fix is macOS-only and cannot be exercised on non-macOS CI. It should be validated on-device using the `leaks`/`heap` stress steps from #238: after repeated failed discovery cycles the `IOHIDManager` and Mach port counts should stay bounded.

## Not included

The secondary `HIDEvent`/`GPProcessMonitor` autorelease cluster (native read-loop pool and #125 revalidation on macOS 26.4.1) is intentionally out of scope here and can follow separately.

Diagnosis and forensics by @maxhis.

### DIFF
--- a/core/hid_gesture.py
+++ b/core/hid_gesture.py
@@ -172,6 +172,15 @@ def _candidate_signature(info) -> dict:
     }
 
 
+def _candidate_cooldown_key(info) -> tuple:
+    """Hashable interface identity for the REPROG_V4-absent cooldown map."""
+    sig = _candidate_signature(info)
+    return (
+        sig["pid"], sig["usage_page"], sig["usage"],
+        sig["transport"], sig["source"], sig["path"],
+    )
+
+
 def _candidate_match_score(info, cached_candidate) -> int:
     """Ranking score for a candidate against the cached identity tuple.
     0 = no match, 1 = PID+usage match, 2 = +same source backend,
@@ -351,6 +360,8 @@ if sys.platform == "darwin":
         _iokit.IOHIDManagerSetDeviceMatching.argtypes = [c_void_p, c_void_p]
         _iokit.IOHIDManagerOpen.argtypes = [c_void_p, c_int]
         _iokit.IOHIDManagerOpen.restype = c_int
+        _iokit.IOHIDManagerClose.argtypes = [c_void_p, c_int]
+        _iokit.IOHIDManagerClose.restype = c_int
         _iokit.IOHIDManagerCopyDevices.argtypes = [c_void_p]
         _iokit.IOHIDManagerCopyDevices.restype = c_void_p
 
@@ -477,6 +488,23 @@ if _MAC_NATIVE_OK:
             finally:
                 _cf.CFRelease(key)
 
+        @staticmethod
+        def _close_manager(manager):
+            """Balance IOHIDManagerOpen before releasing the CF wrapper.
+
+            IOHIDManagerClose releases the IOKit user-client (its Mach ports)
+            and the enclosed-device opens; a bare CFRelease does not, so the
+            manager object and roughly two Mach ports leak on every open. This
+            was the dominant leak in issue #238: tens of thousands of live
+            IOHIDManager objects after a day of reconnect churn."""
+            if not manager:
+                return
+            try:
+                _iokit.IOHIDManagerClose(manager, 0)
+            except Exception:
+                pass
+            _cf.CFRelease(manager)
+
         @classmethod
         def enumerate_infos(cls):
             infos = []
@@ -539,12 +567,24 @@ if _MAC_NATIVE_OK:
                 if matching:
                     _cf.CFRelease(matching)
                 if manager:
-                    _cf.CFRelease(manager)
+                    cls._close_manager(manager)
                 for item in matching_refs:
                     _cf.CFRelease(item)
             return infos
 
         def open(self):
+            """Exception-safe open. On any partial failure, release every
+            CoreFoundation object allocated so far so a failed open cannot
+            leak the manager, matching dict, matching refs, or the retained
+            device. ``close()`` is idempotent and guards each field, so a
+            failed open is safe to unwind through it (issue #238)."""
+            try:
+                self._open()
+            except Exception:
+                self.close()
+                raise
+
+        def _open(self):
             keys = [
                 self._cfstring("VendorID"),
                 self._cfstring("ProductID"),
@@ -641,7 +681,7 @@ if _MAC_NATIVE_OK:
                 _cf.CFRelease(self._matching)
                 self._matching = None
             if self._manager:
-                _cf.CFRelease(self._manager)
+                self._close_manager(self._manager)
                 self._manager = None
             for item in self._matching_refs:
                 _cf.CFRelease(item)
@@ -708,6 +748,15 @@ if _MAC_NATIVE_OK:
                     if deadline is not None:
                         continue
                     return b""
+
+        def __del__(self):
+            # Backstop: an instance discarded without an explicit close() (for
+            # example a caller that drops a half-open device) must not leak
+            # IOKit resources. close() is idempotent; never raise from __del__.
+            try:
+                self.close()
+            except Exception:
+                pass
 
 # ── Constants ─────────────────────────────────────────────────────
 LOGI_VID       = 0x046D
@@ -923,6 +972,12 @@ def _control_present(controls, cid: int) -> bool:
 class HidGestureListener:
     """Background thread: diverts the gesture button and listens via HID++."""
 
+    # Cooldown before an interface that opened but exposed no REPROG_V4 is
+    # re-probed, so a stable set of incompatible interfaces is not re-scanned
+    # every pass during a reconnect storm. Bypassed when it is the only
+    # candidate so a woken device reconnects promptly (issue #238).
+    _REPROG_ABSENT_COOLDOWN_S = 10.0
+
     def __init__(self, on_down=None, on_up=None, on_move=None,
                  on_connect=None, on_disconnect=None, extra_diverts=None,
                  on_wheel=None, on_thumbwheel=None,
@@ -1001,6 +1056,8 @@ class HidGestureListener:
         self._smart_shift_slot_lock = threading.Lock()
         self._smart_shift_event = threading.Event()
         self._reconnect_requested = False
+        # signature -> monotonic deadline; see _REPROG_ABSENT_COOLDOWN_S.
+        self._reprog_absent_until = {}
         self._pending_battery = None
         self._battery_result = None
         self._battery_event = threading.Event()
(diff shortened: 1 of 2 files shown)


mizi:
I've been running this branch for over 12 hours, and memory usage in Activity Monitor stays at 218MB.
leaks says it increased insignificantly, could be just normal memory use:
12 hours ago:
```
Physical footprint:         178.4M
Physical footprint (peak):  178.4M
```
now:
```
Physical footprint:         509.9M
Physical footprint (peak):  749.6M
```
Before these changes, Activity Monitor would already show 2-3GB.

mizi:
20 hours later, it grew more.
```
Physical footprint:         787.6M
Physical footprint (peak):  1.0G
```
I'll try again together with https://github.com/TomBadash/Mouser/pull/242

Issue #253, opened 2026-07-25
[Feature] Will the button function for clicking the middle button of the M590 on both sides be improved?


327607501:
### Feature description

I hope the software can support the function of clicking the buttons on both sides with the middle button of the Logitech M590 mouse. After all, on the M590, these two buttons are physical and support left-clicking with the middle button and right-clicking with the middle button.

<img width="395" height="527" alt="Image" src="https://github.com/user-attachments/assets/63fd156f-0e63-46d7-ab48-54fb7218e741" />

### Use case / motivation

因为这两个是实体的按键，多了两个按键，就代表多了两个自定义的功能，可以提高工作效率。

### Area of the app

Button remapping / actions

### OS relevance

All platforms

### Alternatives considered

It is normal in Logitech Options+, but because Logitech Options+ is too large, I hope that mouser has this function and can completely replace Logitech Options+.

### Device info JSON (if relevant)

```json
{
  "device_key": "m585/m590_multi-device_mouse",
  "display_name": "M585/M590 Multi-Device Mouse",
  "product_id": "0xC52B",
  "product_name": "M585/M590 Multi-Device Mouse",
  "transport": "hidapi",
  "ui_layout": "mx_master_3s",
  "supported_buttons": [
    "middle",
    "gesture",
    "gesture_left",
    "gesture_right",
    "gesture_up",
    "gesture_down",
    "xbutton1",
    "xbutton2",
    "hscroll_left",
    "hscroll_right",
    "mode_shift"
  ],
  "gesture_cids": [
    "0x00D7"
  ],
  "dpi_range": [
    200,
    8000
  ],
  "discovered_features": {
    "REPROG_V4 (0x1B04)": "index 0x0A",
    "BATTERY (0x1000)": "index 0x07"
  },
  "reprog_controls": [
    {
      "index": 0,
      "cid": "0x0050",
      "task": "0x0038",
      "flags": "0x0011",
      "mapped_to": "0x0050",
      "mapping_flags": "0x0000"
    },
    {
      "index": 1,
      "cid": "0x0051",
      "task": "0x0039",
      "flags": "0x0011",
      "mapped_to": "0x0051",
      "mapping_flags": "0x0000"
    },
    {
      "index": 2,
      "cid": "0x0052",
      "task": "0x003A",
      "flags": "0x0171",
      "mapped_to": "0x0052",
      "mapping_flags": "0x0000"
    },
    {
      "index": 3,
      "cid": "0x0053",
[...]

Issue #252, opened 2026-07-25
[Bug] mouser crash when right click


lazyzyf:
### Mouser version

v3.7.1 and all other versions

### How are you running Mouser?

Official GitHub release asset

### Operating system

Windows 11

### Mouse connection type

USB receiver (Bolt / Unifying)

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

Every time when click the right button, Mouser runs very slow, and after one second, it crashes.
i use M575 business.

### Expected behavior

When right-clicking, it does not crash.

### Steps to reproduce

1, click right button
2. it crash

### Device info JSON

```json

```

### Log file contents

```text

```

TomBadash:
Thanks for the report. One thing that may help narrow this down: Mouser doesn't intercept right-click at all — the Windows hook only handles the wheel, middle button and side buttons, so the remap pipeline isn't in the path here. That means something else in the app is reacting to what the right-click *opens*, not to the click itself.

To confirm, could you attach two things?

1. **The log file**: `%APPDATA%\Mouser\logs\mouser.log` — the portion covering a right-click that reproduces the problem.
2. **Where you right-click** when it happens — on the desktop, inside File Explorer, or inside a specific application? This distinction matters a lot for diagnosing it.

Also useful: does the window disappear outright, or does it freeze and Windows offers to close it? And the device info JSON from the app if you can get to it.

My current suspicion is the foreground-app detector: on Windows 11 a shell context menu can send it down an expensive window-scan path that repeats several times a second while the menu is open. If that's what's happening, your log will show repeated `[AppDetect] FG: explorer.exe class=...` lines timed to the click. That's a hypothesis, not a diagnosis — the log will settle it.

lazyzyf:
thanks for your quick reply. 
it happens anytime and anywhere when I do right click. 
as soon as I right click, the mouser cursor begin to freeze and hardly respond, after 1-2 seconds, the mouser.exe process begin to exit automatically.
seems your suspicion is correct, below is the log, any way to fix this?
```
2026-07-25 22:05:29 [HidGesture] DPI set FAILED
2026-07-25 22:05:29 [AppDetect] FG: explorer.exe class=TaskListThumbnailWnd title=''
2026-07-25 22:05:30 [AppDetect] FG: explorer.exe class=TaskListThumbnailWnd title=''
[...]

PR #254, opened 2026-07-26
Fix reversed horizontal wheel-tilt direction on Windows

unijiang:
The WM_MOUSEHWHEEL handler mapped a positive delta to HSCROLL_LEFT, but a positive WM_MOUSEHWHEEL delta means the wheel was tilted to the right (per the Win32 docs). This made hscroll_left/hscroll_right actions fire in the opposite direction on Windows compared to macOS and Linux, whose hooks both map positive -> HSCROLL_RIGHT.

Swap the Windows mapping so positive delta -> HSCROLL_RIGHT and negative -> HSCROLL_LEFT, matching the other platforms. The invert/injection path (which negates delta) is unaffected.

Verified in Windows and macOS.

### DIFF
--- a/core/mouse_hook_windows.py
+++ b/core/mouse_hook_windows.py
@@ -376,12 +376,15 @@ class MouseHook(BaseMouseHook):
 
             elif wParam == WM_MOUSEHWHEEL:
                 delta = hiword(mouse_data)
+                # WM_MOUSEHWHEEL: a positive delta means the wheel was tilted
+                # to the right, negative to the left (per Win32 docs). Match the
+                # macOS/Linux hooks, which also map positive -> HSCROLL_RIGHT.
                 if delta > 0:
-                    event = MouseEvent(MouseEvent.HSCROLL_LEFT, abs(delta))
-                    should_block = MouseEvent.HSCROLL_LEFT in self._blocked_events
-                elif delta < 0:
                     event = MouseEvent(MouseEvent.HSCROLL_RIGHT, abs(delta))
                     should_block = MouseEvent.HSCROLL_RIGHT in self._blocked_events
+                elif delta < 0:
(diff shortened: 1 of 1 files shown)


TomBadash:
Thanks for this — your diagnosis is exactly right, and the Win32 docs plus the macOS/Linux hooks back it up the same way you describe.

I'd landed the same swap on master an hour before this PR was opened, as part of a wider fix for the crash reports in #252 and #253. Rather than let the two paths drift again, the direction decision now lives in a single `hscroll_event_type()` helper in `core/mouse_hook_types.py` that all three hooks call, with tests pinning the sign and a guard that fails if any hook goes back to building the event from a raw delta comparison. Your commit is credited as a co-author on `a701d5b`.

The part I couldn't do myself was your "verified in Windows and macOS" — I could only reason from the docs and the cross-platform inconsistency. Knowing it was confirmed on real hardware is what made me comfortable shipping a change that silently flips existing tilt mappings for anyone who had worked around the old behaviour. Thank you for that, and for the M590 support in #241 — both reporters in #253 are now sorted because of it.

One thing worth checking on your side: this PR was closed two minutes after you opened it, with no comment and a clean merge state. If that wasn't deliberate, it may be that your fork's `device/m590` branch was deleted or reset — GitHub closes a PR automatically when its head branch disappears. Nothing is lost either way, since the fix is on master.

---
_Generated by [Claude Code](https://claude.ai/code)_

unijiang:
@TomBadash , thanks for the thorough response, and glad the diagnosis lined up with what you'd already landed. Appreciate the co-author credit on  `a701d5b` .

On the close: that was deliberate, not a branch mishap. I closed it myself because I saw you already had commits related to this issue on master. I just forgot to leave a message explaining why, so it looked abrupt — sorry about that.

PR #262, opened 2026-07-28
fix: improve transport detection logic for direct HID mouse

### LINKED ISSUE #267: [Bug] Incorrect HID detection
### Mouser version

3.5.3

### How are you running Mouser?

Official GitHub release asset

### Build / asset details

Mouser-macOS.zip

### Operating system

macOS 15 (Sequoia)

### OS version details

MacOS Tahoe 26.6

### Mouse connection type

Bluetooth

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

Mouser shows incorrect HID interface (USB Reciever instead of Bluetooth connection)

Image

### Expected behavior

Show Bluetooth connection

### Steps to reproduce

1. Connect Logitech mouse via bluetooth
2. Open mouser

### Device info JSON

```json

```

### Log file contents

```text

```

### Screenshots or screen recordings

_No response_

### Additional context

_No response_

pincetgore:
fix https://github.com/TomBadash/Mouser/issues/267

If a Bluetooth mouse enumerated on `dev_idx = 1` (or another non-`0xFF` slot), it failed `idx == BT_DEV_IDX` and fell back to `USB Receiver`, even though its Product ID belonged to a standalone Bluetooth mouse rather than a USB dongle.

### Solution

1. Added `UNIFYING_RECEIVER_PID` (`0xC52B`) and a receiver PID check (`0xC5xx` range).
2. Updated transport determination so standalone mouse PIDs (such as `0xB034` / `0xB043`) or Bluetooth candidate interfaces are correctly identified as `"Bluetooth"`.
3. Kept `"Logi Bolt"` for `PID 0xC548` and `"USB Receiver"` for other receiver PIDs (such as `0xC52B`).

### Testing

- Added `test_transport_label_bluetooth_for_direct_mouse_pid` in `tests/test_hid_gesture.py`.
- Ran full test suite: `PYTHONPATH=. pytest tests/test_hid_gesture.py` (50 passed).

### DIFF
--- a/core/hid_gesture.py
+++ b/core/hid_gesture.py
@@ -882,6 +882,7 @@ BT_DEV_IDX     = 0xFF        # device-index for direct Bluetooth
 # Known Logi Bolt receiver PID.
 # Source: https://github.com/pwr-Solaar/Solaar/blob/master/lib/logitech_receiver/base_usb.py
 BOLT_RECEIVER_PID = 0xC548
+UNIFYING_RECEIVER_PID = 0xC52B
 FEAT_IROOT     = 0x0000
 FEAT_REPROG_V4 = 0x1B04      # Reprogrammable Controls V4
 FEAT_ADJ_DPI   = 0x2201      # Adjustable DPI
@@ -3211,12 +3212,30 @@ class HidGestureListener:
                         # firmware does not advertise.
                         self._install_thumb_button_extra(device_spec, controls)
                         self._divert_extras()
-                        if idx == BT_DEV_IDX:
+                        is_receiver_pid = bool(
+                            pid is not None
+                            and (
+                                (int(pid) & 0xFF00) == 0xC500
+                                or int(pid) in (BOLT_RECEIVER_PID, UNIFYING_RECEIVER_PID)
+                            )
+                        )
+                        is_bluetooth = bool(
+                            idx == BT_DEV_IDX
+                            or "bluetooth" in (opened_transport or "").lower()
+                            or "bluetooth" in (candidate_transport or "").lower()
+                            or "ble" in (opened_transport or "").lower()
+                            or "ble" in (candidate_transport or "").lower()
+                            or not is_receiver_pid
+                        )
+
+                        if is_bluetooth:
                             actual_transport = "Bluetooth"
                         elif pid == BOLT_RECEIVER_PID:
                             actual_transport = "Logi Bolt"
-                        else:
+                        elif is_receiver_pid:
                             actual_transport = "USB Receiver"
+                        else:
+                            actual_transport = "Bluetooth"
                         self._connected_device_info = build_connected_device_info(
                             product_id=pid,
                             product_name=hidpp_name or product,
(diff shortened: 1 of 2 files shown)


pincetgore:
I've just refined transport detection logic (Bluetooth / USB Receiver / Logi Bolt)

### Context
In `HidGestureListener._try_connect()`, the transport label used to be derived solely from `dev_idx`: `idx == BT_DEV_IDX → Bluetooth`, otherwise → receiver. This produced a false `"USB Receiver"` label whenever a directly-connected (BT) device happened to answer the feature probe on a receiver slot (`1..6`) instead of `0xFF` — a side effect of the `idx_order` fallback, which for the "unknown" case tries `(BT_DEV_IDX, 1, 2, ..., 6)`.

### What changed
- Added a PID-range check: any `pid` in `0xC500–0xC5FF` (via `RECEIVER_PID_MASK`/`RECEIVER_PID_PREFIX`) identifies a Logitech receiver (Unifying, Bolt, etc.); any other PID is treated as a direct connection.
- Final rule: if the PID is **not** in the receiver range, the transport is always reported as `"Bluetooth"` (the app officially supports only Bluetooth and USB-receiver transports, per the README). Explicit BT signals (`dev_idx == BT_DEV_IDX`, or `"bluetooth"`/`"ble"` in `opened_transport`/`candidate_transport`) still override a receiver-range PID for the rare edge case where a direct device answers on a receiver slot.
- Removed the redundant explicit check `pid in (BOLT_RECEIVER_PID, UNIFYING_RECEIVER_PID)` — the `0xC500–0xC5FF` mask already covers both PIDs, so the explicit list added nothing.
- Removed the unreachable `else: actual_transport = "Bluetooth"` branch in the final `if/elif` chain. It was dead code: the prior condition already forced `"Bluetooth"` for every non-receiver PID via `or not is_receiver_pid`, so that branch could never execute.
- Renamed `is_bluetooth` to `has_bt_signal` to clearly separate "BT evidence from transport string/index" from the final transport decision.
- Added comments explaining both rules (the mask already covers the named constants; a non-receiver PID defaults to Bluetooth).

### Tests
- Added `test_transport_label_bluetooth_for_direct_mouse_pid`: a direct device (MX Master 3S PID `0xB034`) that answers on `dev_idx=1` must be labeled `"Bluetooth"`, not `"USB Receiver"`.
- Existing tests `test_transport_label_logi_bolt_for_bolt_receiver` and `test_transport_label_usb_receiver_for_non_bolt` (Unifying `0xC52B`) still pass unchanged — receiver-PID behavior is untouched.
- `pytest tests/test_hid_gesture.py` — 50/50 passed.

### Deliberately not changed
Any non-receiver PID is unconditionally treated as Bluetooth, even in the hypothetical case of a directly wired USB connection (e.g. G502 over a USB cable). This is already the app's stated scope (README lists only Bluetooth and USB-receiver support), and G502-family devices don't expose `REPROG_CONTROLS_V4` anyway, so they never reach this code path — no actual regression today, but worth keeping in mind if a wired HID++2.0 mouse is ever added to the catalog.


PR is ready for review

Issue #47, opened 2026-03-24
M720 Triathlon Support


LagLang:
Is it possible to add support for the M720 Triathlon?

SeSamSeg:
Apparently the gesture button is 0x00D0.
With the following config most buttons seem to work:
```
    LogiDeviceSpec(
        key="m720_triathlon",
        display_name="M720 Triathlon",
        product_ids=(0xB015,),
        aliases=("M720 Triathlon Multi-Device Mouse",),
        ui_layout="mx_master",
        image_asset="icons/mouse-simple.svg",
        supported_buttons=MX_MASTER_BUTTONS,
        gesture_cids=(0x00D0,),
        dpi_min=200,
        dpi_max=8000,
    ),
```

Notes:
1. Horizontal scroll: left/right buttons work, although they seem flipped as moving the scroll wheel to the left, triggers 
2. Mode shift: remap does not work. I don't think it's any of the reprog_controls are for that button
    I did notice that you get a (255, 30, 1, 0, [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]) and (255, 30, 1, 0, [1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]) message on switching scroll modes
    The following messages are in the log:
```
    [HidGesture] HID++ error 0x02 (INVALID_ARGUMENT) for feat=0x0B func=0x3 devIdx=0xFF req=[00 C4 03 00 00] resp=[3A 02 00 00 00 00 00 00 00 00 00 00 00 00 00 00]
    [HidGesture] Extra divert 0x00C4 (Smart Shift): FAILED
```
3. Haven't tested dpi switching

PR #274, opened 2026-08-10
feat(device): Add M720 triathlon support

Las-re:
Add Logitech M720 Triathlon / Precision Pro mouse support as requested by #47, #199 

Uses an edited CC BY 2.0 Image instead of a product render to be extra safe.

- Added M720_BUTTONS tuple
- Added M720 Catalog spec with comments. Product IDs might be under-specified, I'm doing all of this  with a Precision Pro and not a _true_ M720.
- Added hotspots.
- Edited existing M720 tests. I don't have a unifying receiver to test with, hoping the PID is correct.
- Gesture control support!

Further work:

1. Battery levels seem to conflict from Bluetooth menu reporting.
2. CREDITS.txt may not be needed, included to be safe. 
3. Replace image with better one to fit theming, polish.


Las-re:
force-pushed to fix timing, sorry.

Also I've treated the DPI as fixed since these mice use an old fixed-DPI sensor.

Issue #287, opened 2026-08-19
macOS wheel-invert tests: 4 stale tests patch a dead method (skipped on CI), + a flaky battery-poll test


davidnoyes:
## Summary

Two macOS-only test-suite problems, invisible on CI because the Ubuntu runner has no PyObjC and skips `MacOSSuppressionTests` (and the relevant class `setUp` `skipTest`s on import failure).

### 1. Four stale tests patch a dead method

`MacOSSuppressionTests` asserts on `_negate_scroll_axis` — an in-place negation helper that the event-tap callback **never calls** (`git log -S "_negate_scroll_axis(cg_event"` returns nothing). The OS-fallback scroll inversion actually runs through `_post_inverted_scroll_event` (block-and-reinject). So on macOS these 4 tests fail:

- `test_os_inversion_runs_when_native_inactive`
- `test_horizontal_inversion_negates_axis_2_in_place`
- `test_both_axes_inverted_in_single_pass`
- `test_os_inversion_resumes_when_logitech_reconnects`

Fixed in #288 (repoint to `_post_inverted_scroll_event`, remove the dead helper + its unit test).

**Correction to my original report:** I initially framed this as a user-facing bug — that the OS fallback inverts non-Logitech scroll when no Logitech is bound. **On-device testing disproved that.** `_should_intercept_events()` (`True` iff `_connected_device is not None`) already makes the whole tap a pass-through when no Logitech is connected, so a generic mouse's scroll is never touched. There is no device-scoping bug; the tests just patch the wrong method.

### 2. A nondeterministic battery-poll test

`test_engine.EngineReplayPhaseOneTests.test_battery_poll_skips_smart_shift_reads_while_replay_is_inflight` passes in isolation but **fails intermittently** in the full-suite run on `master` (observed ~1 in 2–3 runs). Looks like a threading/test-isolation race. Not addressed by #288 — flagged here for separate cleanup.

### Suggestion

Consider running the macOS test suite in a macOS CI job (or installing PyObjC) so `MacOSSuppressionTests` stops being silently skipped.

PR #288, opened 2026-08-19
test(macos): fix stale wheel-invert tests, drop dead scroll helper

davidnoyes:
## Summary

Refs #287. **Test-hygiene only — no behaviour change.**

`MacOSSuppressionTests` patched `_negate_scroll_axis`, a dead method the macOS event-tap callback never calls. The OS-fallback scroll inversion actually runs through `_post_inverted_scroll_event` (block-and-reinject). CI runs on Ubuntu (no PyObjC), so these tests are skipped there and the 4 failures were only visible on macOS.

## Changes

- Repoint the 4 fallback tests at `_post_inverted_scroll_event` with reinject semantics (a single call carries both axes; the original event is blocked and an inverted copy reinjected).
- Remove the dead `_negate_scroll_axis` helper and its direct unit test.
- Keep the "no Logitech" / "reconnect" gate tests unchanged — the device scoping they check is already provided by `_should_intercept_events()` (`True` iff `_connected_device is not None`), which makes the whole tap a pass-through when no Logitech is bound.

## Testing

- `test_wheel_divert`: 39 tests green on macOS (was 4 failing). Rest of the suite unchanged.
- Verified on-device: with no Logitech connected, scroll from a generic mouse passes through untouched — confirming `_should_intercept_events()` already scopes inversion to a connected Logitech, so no behavioural gate was needed.

## Note

While verifying I found `test_engine…test_battery_poll_skips_smart_shift_reads_while_replay_is_inflight` fails **nondeterministically** in the full suite on `master` too (a threading/isolation flake, unrelated to this change). Flagged in #287 as separate cleanup.

### DIFF
--- a/core/mouse_hook_macos.py
+++ b/core/mouse_hook_macos.py
@@ -77,19 +77,6 @@ class MouseHook(BaseMouseHook):
         self._dispatch_thread = None
         self._first_event_logged = False
 
-    def _negate_scroll_axis(self, cg_event, axis):
-        for field_name in (
-            f"kCGScrollWheelEventDeltaAxis{axis}",
-            f"kCGScrollWheelEventFixedPtDeltaAxis{axis}",
-            f"kCGScrollWheelEventPointDeltaAxis{axis}",
-        ):
-            field = getattr(Quartz, field_name, None)
-            if field is None:
-                continue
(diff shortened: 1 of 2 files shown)


davidnoyes:
Auto-closed when the branch was renamed to match the test-only scope. Continued in #289.

PR #264, opened 2026-07-28
fix(windows): recover the hook and HID++ handle after sleep

TomBadash:
Fixes #263. Also covers #257, #246 and #258, which are the same failure without the sleep.

## The problem

Mouser stops working entirely after a Windows resume and stays that way until it is restarted. Two independent failures can put it there, and **neither one raises anything**:

**The low-level hook.** Windows removes a `WH_MOUSE_LL` hook whose procedure overruns `LowLevelHooksTimeout` (300 ms by default) and tells nobody — the `HHOOK` stays non-NULL and simply never fires again. Resume is the classic trigger (the process is still being scheduled back in while a burst of input arrives), but it also happens under load, which is the un-slept variant in #257 / #246. The only thing that reinstalled it was `WM_DEVICECHANGE`, which does not necessarily fire on Modern Standby where USB stays enumerated.

**The HID++ handle.** A resume can leave it open against a device stack Windows has already replaced. `_rx()` returns `None` on timeout and raises only on a real error, so reads never fail — they just never return anything again, and the listener sits `_connected` against a dead pipe.

**Why nothing caught it.** The listener already has a full escalation for this (`_consecutive_request_timeouts` → device-asleep → 300 s → reconnect), but it only advances off *failed HID++ requests*, and `engine._background_hid_poll_allowed` gates all background traffic on `_frontend_visible`, which is `False` whenever the window is hidden or minimised — i.e. exactly the state Mouser is left in over a sleep. So no requests, no timeouts, no detection. This is also why #258 reports that *opening the UI* fixes it: that re-enables polling, the first request fails, and the existing recovery chain finally runs.

And the two failures compound: when HID++ does notice a disconnect, `_connected_device` goes `None` and `_should_intercept_events()` deliberately turns the hook into a pass-through, so one dead channel silences the other.

macOS already rebuilds on `NSWorkspaceDidWakeNotification` (`mouse_hook_macos.py`). Windows had no equivalent at all — this is a platform gap more than a deep bug. The reporter's workaround (a Task Scheduler job on resume event 107) works because it does by hand exactly what the code never did.

## The fix

**1. Windows learns that the machine woke.** The existing hidden message window now subscribes to `WM_POWERBROADCAST`, console display-state (`GUID_CONSOLE_DISPLAY_STATE` — required, since Modern Standby does not reliably raise the APM resume messages) and session unlock. All of them funnel into one recovery pass; a burst of cues for a single wake collapses via a 5 s dedupe.

[...]

Issue #258, opened 2026-07-27
[Bug] MX Master 3 HID++ configuration is lost after mouse idle while Mouser remains running


zhuwei3310:
### Mouser version

3.7.1

### How are you running Mouser?

Official GitHub release asset

### Operating system

Other

### OS version details

macOS 26.5.2 Apple Silicon Mac

### Mouse connection type

Bluetooth

### Logitech Options+ status

- [x] I have confirmed that Logitech Options+ is **not** running

### What happened?

When Mouser runs in the background, after the mouse is unused for several minutes, the mouse stops using Mouser's configured HID++ settings and silently falls back to the mouse's default hardware behavior.

The computer remains fully active during reproduction:

- No macOS sleep.
- No display sleep.
- No system lock.
- No Bluetooth reconnect.

Only the mouse is idle.

Mouser itself continues running normally in the background (menu bar icon remains available), but the mouse behaves as if Mouser is no longer controlling the device.

Opening the Mouser UI immediately triggers a reconnect, after which all configurations are restored.

### Expected behavior

Mouser should detect that the HID++ connection or device configuration has become invalid while running in the background.

- It should automatically:
- reconnect to the mouse;
- restore HID++ configuration;

reapply gesture buttons, Smart Shift, wheel settings, and DPI.

Opening the UI should not be required.

### Steps to reproduce

1. Start Mouser 3.7.1.
2. Connect Logitech MX Master 3 through Bluetooth.
3. Confirm that Mouser configuration is active:
   -   Gesture buttons are remapped.
   -   Smart Shift settings are applied.
   -   Scroll inversion settings are active.
   -   DPI configuration is applied.
4. Keep using the computer normally, but do not move or click the mouse for around 5 minutes.
5. Move or click the mouse again.

### Device info JSON

```json
{
  "device_key": "mx_master_3",
  "display_name": "MX Master 3",
  "product_id": "0xB023",
  "product_name": "Wireless Mouse MX Master 3",
  "transport": "Bluetooth",
  "ui_layout": "mx_master_3",
  "supported_buttons": [
    "middle",
    "gesture",
    "gesture_left",
    "gesture_right",
    "gesture_up",
    "gesture_down",
    "xbutton1",
    "xbutton2",
    "hscroll_left",
    "hscroll_right",
    "mode_shift"
  ],
  "gesture_cids": [
    "0x00C3",
    "0x00D7"
  ],
  "dpi_range": [
    200,
    8000
  ],
[...]

PR #292, opened 2026-08-22
fix(hid): reapply the wheel native-invert when the device wakes from idle sleep

vvoody-:
Refs #258.

## The problem

Inverted horizontal scroll silently stops working once the mouse has been idle for a while. The toggle in the UI still reads *on*, and scrolling stays wrong until the setting is toggled or Mouser is restarted.

**The setting is firmware state, and it is volatile.** Mouser does not invert scroll in software when it can avoid it: `_apply_wheel_invert_setting` writes the invert bit into the device itself via `0x2150` (thumbwheel) and `0x2121` (hi-res wheel), so the OS receives native scroll with the direction already flipped and the scroll cadence survives end to end. An MX Master parks itself at the end of a long idle, and that power cycle clears those bits.

**Nothing told Mouser to write them again.** The device never disconnects. The receiver holds the HID handle open across the whole sleep/wake cycle, so there is no disconnect, no reconnect, and no read error — every path that could have replayed the write is simply unreachable. The listener's own sleep detector cannot help either: it only advances off *failed HID++ requests*, and `engine._background_hid_poll_allowed` gates all background traffic on `_frontend_visible`, which is `False` whenever the window is hidden or minimised — exactly how Mouser is left running over a long idle. No requests, no timeouts, no detection.

**And the engine would have skipped the write even if something had fired.** `_apply_wheel_invert_setting` has a fast path that returns early when nothing it tracks has changed. After a device-side power cycle nothing on Mouser's side *has* changed — config, UI toggle, `_wheel_divert_active_local` and `_last_native_invert_target` all still say "inverted". So the write is skipped. Worse, `hook.wheel_native_invert_active` stays `True`, which is what tells the OS-layer inversion in `mouse_hook_windows.py` / `_linux` / `_macos` to stand down. Both inversion paths end up off at once, which is why the symptom is a total loss of inversion rather than a fallback to the software path.

## The evidence

The device is not silent about this — it announces it, and Mouser was not listening. Captured from an MX Master 4 on a Bolt receiver:

```
rx 11 02 04 00 01 01 01 ...
      ^^ ^^    ^^ ^^ ^^
      |  |     |  |  └─ reason:  power switch activated
      |  |     |  └──── request: software reconfiguration needed
      |  |     └─────── status:  reconnection
      |  └───────────── feature index 0x04, event func 0
      └──────────────── device index 0x02
```

That is `0x1D4B` **Wireless Device Status**, the purpose-built "I re-linked, reconfigure me" broadcast. Mouser never resolves the feature, so the report fell through `_on_report` unhandled.

Two runs bracket the failure exactly. Before the change, a full power-switch cycle produced **that one line and nothing else** — no `Device Connected`, no reconnect, no replay, and the mouse scrolled the wrong way afterwards. After the change, the same bytes drive a replay in the same second.

The idle case — the one users actually hit — was then confirmed on hardware: the mouse was left untouched until it parked itself, and woken by movement alone with the power switch never touched.

```
07:10:11  (last activity)
          ... 2 h 29 min of total HID++ silence, handle open the whole time ...
09:40:11  rx 11 02 04 00 01 01 01 ...
09:40:11  Device re-linked (0x1D4B status=0x01 request=0x01 reason=0x01)
09:40:11  Device wake — replaying wheel native-invert
09:40:11  Wheel native invert v=False h=True vertical=OK thumb=OK
```

Horizontal scroll direction confirmed correct after the wake. Across multi-hour sessions the cycle has since fired six times — every one decoded, every firmware write acknowledged, no read errors or tracebacks.

One trap worth recording, and the reason there is a test pinning it: the `reason` byte reads `0x01` (*"power switch activated"*) for a natural idle wake too, not just a switch cycle. Filtering on it looks correct against the spec and would silently drop every idle wake — i.e. the entire bug.

## The fix

**1. Hear the wake.** `hid_gesture` resolves `0x1D4B` at connect alongside the other optional features and treats its `statusBroadcastEvent` as a wake signal. Either flag is enough — *reconnection* says the link came back, *reconfiguration needed* is the device asking outright. An event carrying neither is something else and stays quiet. A device without the feature falls back to the old reconnect-only behaviour.

The absent branch *clears* the index rather than only logging it. Feature resolution runs per receiver slot, but the reset runs once per interface candidate, so a slot that resolves `0x1D4B` and then fails to divert would leave its index behind for the next slot. This branch sits in front of the whole dispatch in `_on_report`, so a stale index colliding on the next slot would swallow every button event.

**2. Route it out of the listener.** `BaseMouseHook.set_device_wake_callback` is deliberately separate from the connection-change callback: the handle never dropped and `device_connected` never flipped, so this is not a connection change and must not be modelled as one.

**3. Write it, without asking the cache.** The engine reapplies with `force=True`, bypassing the stale fast path. The write is deferred to a worker via `_schedule_wheel_invert_apply` — the callback arrives on the HID listener thread and the write blocks until that same thread services it, so doing it inline would deadlock until the request times out.

The connect path now forces too. After a reconnect the firmware state is unknown regardless of what the engine has cached, so the fast path was wrong there as well.

The write stays with the engine rather than being fired from the listener on purpose: the engine is the side that knows the config, and it is the side that must clear `wheel_native_invert_active` if the write fails, so the OS-layer fallback takes over instead of leaving no inversion at all.

**4. One worker, and no dropped wakes.** Only one apply runs at a time — a burst would otherwise stack threads that each sit on `_wheel_divert_call_lock` for the full timeout. A request arriving while one is in flight is *coalesced into a re-run rather than dropped*, mirroring `_request_saved_settings_replay`: a wake means the firmware forgot its state, so discarding one is the exact failure this path exists to prevent, and the in-flight write may already have landed before the device power-cycled again.

The flag handling is spelled out in the comments; the short version is that no request can be lost between the pending check and the flag clear, and a worker that never starts wedges nothing.

## Scope — what this does not do

The wake hook is generic, but only the wheel native-invert is wired to it, because that is the setting verified to be volatile on real hardware. The diverts already set the persistent HID++ bit (`0x02`) and survive sleep by design, and DPI / Smart Shift were not observed to be lost in these runs. If reports show other settings dropping on wake, they belong on this same hook rather than a new mechanism.

Verified on **MX Master 4 / Logi Bolt / Windows 11**. #258 is an MX Master 3 over Bluetooth on macOS — `0x1D4B` is a standard feature and the mechanism is transport-agnostic, but I have no capture from that setup, so this is deliberately filed as `Refs` rather than `Fixes`. Happy to change it if a reporter confirms.

## Relationship to #264

This complements #264 rather than competing with it. Both restore a broken device connection, but they catch genuinely different failures, and neither subsumes the other.

| | #264 | this PR |
|---|---|---|
| What breaks | HID handle goes stale — reads never return | Handle is healthy; the **device forgets its volatile settings** |
| Detection | Liveness probe fails → reconnect | Device announces it via `0x1D4B` |
| Trigger | System resume, or a dropped low-level hook | The mouse's own idle park |

#264's watchdog reconnects only when its root request goes unanswered twice; on a wake like the one captured above the probe *succeeds*, because the handle was never dead — so it returns without replaying anything. That is by design, and its comments say so: it is built to act only on independent evidence that the channel is dead. This case has no such evidence to offer.

They also do not collide. #264 does not touch `engine.py`, contains no reference to `0x1D4B`, and its `hid_gesture.py` changes sit in different methods.

## Tests

`tests/test_wheel_divert.py`, built around the captured payload as the reference case. Decoding: each flag combination, our own polled responses, reports arriving before the feature is resolved, and the `reason`-byte trap above. Replay: the engine writes *despite* every cached value being unchanged — the exact condition the fast path used to swallow. Concurrency: coalescing, one worker per burst, and neither a raising apply nor a worker that fails to start losing a wake or wedging the scheduler.

The thread-based tests use `Event` handshakes rather than sleeps, so they are deterministic. Full suite passes with no new failures.

---

🤖 Authored with [Claude Code](https://claude.com/claude-code) (Claude Opus 5).

Every capture above is from a live MX Master 4 on a Logi Bolt receiver — observed, not inferred. The byte dumps came from temporary local instrumentation, not from anything in this diff.


Sada-Mio:
Tested a merge of #292 and #264 on Windows 11:

Wake from PC sleep: Resolved. 
Gesture button assigned action of MX Master 3 persist as expected.

Power cycle: Still an issue. (physical ON/OFF switch)
Deep Sleep: Still an issue.  (10-15min idle deep sleep)
Both requires relaunch of Mouser.exe

